PR #5, opened 2023-09-07
Fix autobind

goccy:
- Fix handling of Used flag when Autobind is enabled
- Automatically assigns a generated name when the name is empty
- Add name validation
- Add an example for autobind's behavior


resolver/resolver.go
```diff
@@ -662,12 +663,30 @@ func (r *Resolver) resolveAutoBindFields(ctx *context, msg *Message) {
...
 }
 
+var nameRe = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9_]*$`)
+
+func (r *Resolver) validateName(name string) error {
+	if name == "" {
+		return nil
+	}
+	if !nameRe.MatchString(name) {
+		return fmt.Errorf(`"%s" is invalid name. name should be in the following pattern: ^[a-zA-Z][a-ZA-Z0-9_]+$`, name)
```
kazegusuri, line 685: This is a different pattern from the regex defined above. Is this intentional? `*` => `+`

kazegusuri (approved):

resolver/resolver.go
goccy: Oh, It’s mistakes! I’ll fix it. Thank you 😄

kazegusuri (approved):

PR #15, opened 2023-09-22
Improve graph cycle detection

shuheiktgw:
## NOTE

The current cycle detection algorithm is computationally expensive, but IIUC, all we have to do is quite simple and check if we can write a path without referencing the same node 🙂  I've introduced a new variable (`visited`) to memorize the path to the current node to make it more effective (O(N)).

### DIFF
--- a/resolver/context.go
+++ b/resolver/context.go
@@ -1,15 +1,15 @@
 package resolver
 
 type context struct {
-	errorBuilder *errorBuilder
-	allWarnings  *allWarnings
-	fileRef      *File
-	svc          *Service
-	mtd          *Method
-	msg          *Message
-	enum         *Enum
-	depIdx       int
-	argIdx       int
+	allErrors   *Error
+	allWarnings *allWarnings
+	fileRef     *File
+	svc         *Service
+	mtd         *Method
+	msg         *Message
+	enum        *Enum
+	depIdx      int
+	argIdx      int
 }
 
 type allWarnings struct {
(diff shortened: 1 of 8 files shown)


shuheiktgw:
No, the current algorithm is wrong. Let me think about it a little more deeply 🤦‍♂️

resolver/resolver.go
```diff
@@ -82,7 +82,8 @@ func (r *Resolver) Resolve() (*Result, error) {
 	// In order to return multiple errors with source code location information,
 	// we add all errors to the context when they occur.
 	// Therefore, functions called from Resolve() do not return errors directly.
-	// Instead, it must return all errors captured by context in ctx.error().
+	// Instead, it must return all errors captured by context in ctx.error()
+	// except for those that fail subsequent processing, such as cyclic dependencies
```
shuheiktgw, line 86: As I commented here, I've changed the r.resolveMessageArgumentReference method and made it return an error. That is because if we call r.servicesWithRule with cyclic dependencies in MessageDependencyGraph, it causes a stack overflow since the method parses the graph as well. So I needed to stop it somehow.

shuheiktgw, line 86: As I commented here, I've changed the r.resolveMessageArgumentReference method and made it return an error. That is because if we call r.servicesWithRule with cyclic dependencies in MessageDependencyGraph, it causes a stack overflow since the method parses the graph as well. So I needed to stop it somehow.

resolver/context.go
```diff
@@ -126,7 +126,7 @@ func (c *context) argIndex() int {
 	return c.argIdx
 }
 
-func (c *context) addError(e *LocationError) {
```
shuheiktgw, line 129: For https://github.com/mercari/grpc-federation/pull/15/files#diff-732f365bd38694dea48332c13814847ee19a76bebc5f3d2c87733f1a4ee915ffR99

resolver/resolver_test.go
```diff
@@ -2140,6 +2141,40 @@ func TestLiteral(t *testing.T) {
 	}
 }
 
+func TestValidation(t *testing.T) {
```
shuheiktgw, line 2144: The validations themselves are implemented in the resolver.go, so I put the test here. Also, it seems invalid proto files are in the validator/testdata directory, so I put the test data files there.

resolver/resolver.go
```diff
@@ -192,6 +193,14 @@ func (r *Resolver) servicesWithRule(ctx *context, files []*File) []*Service {
 }
 
 func (r *Resolver) validateServiceDependency(ctx *context, service *Service) {
+	// Avoid an infinite loop caused by the cyclic dependency
```
shuheiktgw, line 196: We must stop here since if we proceed; it may go into an infinite loop if the graph contains cyclic dependency. This is why I introduced `CyclicDependencyError` in this PR.

resolver/error.go
```diff
@@ -22,13 +22,52 @@ func (e *Error) Error() string {
...
+
+func (e *LocationErrorImpl) GetMessage() string {
+	return e.Message
+}
+
+func (e *LocationErrorImpl) Error() string {
+	return e.Message
+}
+
+type CyclicDependencyError struct {
```
shuheiktgw, line 57: We may be able to contain more valuable information for users in case of cyclic dependency such as dependency path.

goccy:
I try to create the same fixing based this PR #16 . Please see it.

shuheiktgw:
Closing in response to https://github.com/mercari/grpc-federation/pull/17

PR #17, opened 2023-09-27
Fix cyclic dependency issues

shuheiktgw:
Created a new PR to fix the problem tackled in https://github.com/mercari/grpc-federation/pull/15 and https://github.com/mercari/grpc-federation/pull/16.

### DIFF
--- a/resolver/graph.go
+++ b/resolver/graph.go
@@ -398,7 +398,7 @@ func validateMessageRuleNodeCyclicDependency(target *MessageRuleDependencyGraphN
 			}
 		}
 		return ErrWithLocation(
-			fmt.Sprintf(`found cyclic dependency for "%s.%s" message`, target.Message.PackageName(), target.Message.Name),
+			fmt.Sprintf(`found cyclic dependency for "%s.%s" message. dependency path: %s`, target.Message.PackageName(), target.Message.Name, dependencyPath),
 			source.MessageLocation(target.Message.File.Name, target.Message.Name),
 		)
 	}
--- a/resolver/graph.go
+++ b/resolver/graph.go
@@ -3,6 +3,7 @@ package resolver
 import (
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/mercari/grpc-federation/source"
 )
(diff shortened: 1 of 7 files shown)


resolver/graph.go
```diff
@@ -370,29 +368,51 @@ func validateMessageRuleGraph(graph *MessageRuleDependencyGraph) error {
...
-		if err := validateMessageRuleNodeCyclicDependency(target, child.Children); err != nil {
-			return err
+		dependencyPath := strings.Join(messages, " => ")
+
+		msg := target.BaseMessage
+		for idx, dep := range msg.Rule.MessageDependencies {
+			if dep.Message == target.Message {
+				return ErrWithLocation(
+					fmt.Sprintf(
+						`found cyclic dependency for "%s.%s" message in "%s.%s. dependency path: %s"`,
```
shuheiktgw, line 392: Added dependency path for users (the same as validateMessageNodeCyclicDependency). Note unlike `visited` we don't have to pop an element from `path` since `append` won't change the slice itself.

resolver/service.go
```diff
@@ -144,7 +144,12 @@ func (s *Service) UseServices() []*Service {
 	return services
 }
 
-func (s *Service) useServices(resolver *MessageResolver) []*Service {
+func (s *Service) useServices(resolver *MessageResolver, msgResolverMap map[*MessageResolver]struct{}) []*Service {
```
shuheiktgw, line 147: Slightly updated from https://github.com/mercari/grpc-federation/pull/16 and passed `msgResolverMap` to the `useServices` method directly so that we don't have to define the `useServicesRecursive` method.

resolver/graph.go
```diff
@@ -370,29 +368,51 @@ func validateMessageRuleGraph(graph *MessageRuleDependencyGraph) error {
...
+						target.Message.PackageName(), target.Message.Name,
+						msg.PackageName(), msg.Name, dependencyPath,
+					),
+					source.MessageDependencyMessageLocation(msg.File.Name, msg.Name, idx),
+				)
+			}
 		}
-		if err := validateMessageRuleNodeCyclicDependency(child, child.Children); err != nil {
+		return ErrWithLocation(
```
goccy, line 400: I think this path isn't actually evaluated, but I would like to display the `depedencyPath` here as well .

shuheiktgw: Ah, I was going to add it there as well but totally forgot it... 🤦‍♂️

goccy (approved):

goccy:
Thank you for your great PR !!! LGTM !

PR #22, opened 2023-09-28
Set up octocov to report coverage

shuheiktgw:
SSIA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ install/vscode-dependencies:
 
 .PHONY: test
 test: test/examples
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -96,5 +96,4 @@ test/examples/%:
 
 .PHONY: cover-html
 cover-html:
```
goccy, line 98: Since the `cover-html` task did two jobs ( `go test -cover` and display with html ), so it seems necessary to run test as well

goccy (approved):

goccy:
LGTM !!!

PR #25, opened 2023-10-03
Fix building of resolver group for same message arguments

goccy:
Fixed a problem in which a resolver group could not be created properly when the same message argument was specified in multiple messages.

### Example

```proto
message GetResponse {
  option (grpc.federation.message) = {
    messages { name: "uid", message: "UserID" }
    messages { name: "user", message: "User", args { name: "user_id", by: "uid.value" } }
    messages { name: "user2", message: "User", args { name: "user_id", by: "uid.value" } }
  };
  User user = 1 [(grpc.federation.field).by = "user"];
  User user2 = 2 [(grpc.federation.field).by = "user2"];
}
```

ResolverGroups are the following.

### BEFORE

```
	   uid ─┐
	         user ─┐
	        user2 ─┤
```

### AFTER

```
	   uid ─┐
	         user ─┐
	   uid ─┐      │
	        user2 ─┤
```


resolver/graph.go
```diff
@@ -337,23 +335,15 @@ func CreateMessageRuleDependencyGraph(ctx *context, baseMsg *Message, rule *Mess
 		}
 	}
 
-	nodeMap := make(map[*MessageRuleDependencyGraphNode]struct{})
-	for _, node := range msgToNode {
-		nodeMap[node] = struct{}{}
-	}
-	for _, node := range nameToNode {
-		nodeMap[node] = struct{}{}
+	allNodes := make([]*MessageRuleDependencyGraphNode, 0, len(msgToNodes))
```
shuheiktgw, line 338: Would it be computationally cheaper if we only sort the roots? Currently, it sorts `allNodes` but IIUC, we only want sorted roots, so what if we first get roots and sort them instead?

```
	var roots []*MessageRuleDependencyGraphNode
	for _, nodes := range msgToNodes {
		for _, node := range nodes {
			if len(node.Parent) == 0 {
				roots = append(roots, node)
			}
		}
	}
	sort.Slice(roots, func(i, j int) bool {
		return roots[i].FQDN() < roots[j].FQDN()
	})
```

shuheiktgw (approved):
LGTM 👍

PR #26, opened 2023-10-03
Add tests for the FindLocationByPos method

shuheiktgw:
SSIA. I've added missing tests for the FindLocationByPos method

### DIFF
--- a/source/file.go
+++ b/source/file.go
@@ -305,6 +305,18 @@ func (f *File) findMessageArgumentByPos(ctx findContext, pos Position, list []*a
 		for _, arg := range literal.Elements {
 			fieldName := arg.Name.Name.AsIdentifier()
 			switch fieldName {
+			case "name":
+				value, ok := arg.Val.(*ast.StringLiteralNode)
+				if !ok {
+					return nil
+				}
+				if f.containsPos(value, pos) {
+					ctx.messageDepOption.Args = &ArgumentOption{
+						Idx:  idx,
+						Name: true,
+					}
+					return f.buildLocation(ctx)
+				}
 			case "by":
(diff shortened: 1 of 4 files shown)


source/file.go
```diff
@@ -305,6 +305,18 @@ func (f *File) findMessageArgumentByPos(ctx findContext, pos Position, list []*a
 		for _, arg := range literal.Elements {
 			fieldName := arg.Name.Name.AsIdentifier()
 			switch fieldName {
+			case "name":
```
shuheiktgw, line 308: It seems the name field is missing, so I've added the support 🙂

source/file_test.go
```diff
@@ -0,0 +1,348 @@
...
+func TestFile_FindLocationByPos(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		desc     string
+		pos      source.Position
+		expected *source.Location
+	}{
+		{
+			desc: "find nothing",
```
goccy, line 22: Currently, `testdata/service.proto` is read directly from each tests. So it may be possible to pass the file name and read that file. In this case, the `testdata` prefix should be omitted and `service.proto` or `service` would be fine.

source/file_test.go
```diff
@@ -0,0 +1,348 @@
...
+			expected: nil,
+		},
+		{
+			desc: "find a ServiceDependencyOption name",
+			pos: source.Position{
+				Line: 15,
+				Col:  15,
+			},
+			expected: &source.Location{
+				FileName: filepath.Base("testdata/service.proto"),
```
goccy, line 36: It seems OK to specify `"service.proto"` directly

goccy (approved):
LGTM !

PR #30, opened 2023-10-10
Fix errgroup context name conflicts

shuheiktgw:
## NOTE
Currently, the context returned from errgroup.WithContext overrides the existing `ctx` variable, and as a result, the `ctx` is canceled after calling `eg.Wait()`, and users receive an error. To fix that problem, I've renamed the context returned from errgroup.WithContext as `egCtx` and pass the name to child templates.


generator/code_generator.go
```diff
@@ -1877,7 +1877,11 @@ func loadTemplate(path string) (*template.Template, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to read template: %w", err)
 	}
-	tmpl, err := template.New("").Parse(string(tmplContent))
+	tmpl, err := template.New("").Funcs(
+		map[string]any{
+			"map": CreateMap,
```
shuheiktgw, line 1882: The builtin `template` function can receive only one argument, but to pass the name of the context, we somehow need to pass multiple arguments. To circumvent the problem, I created a function to create a map.

goccy (approved):

goccy:
LGTM !

PR #31, opened 2023-10-11
support CEL for "by" feature

goccy:
Support [CEL](https://github.com/google/cel-spec) for "by" feature.

Also, I've refactored resolver's process. 
- replace `resolver.Value.Literal` with `resolver.Value.Const`
- add `resolver.Value.CEL` to store CEL value


_examples/01_minimum/federation/federationservice_grpc_federation.go
```diff
@@ -207,6 +349,30 @@ func recoverErrorFederationService(v interface{}, rawStack []byte) *FederationSe
 	}
 }
 
+func (s *FederationService) evalCEL(expr string, vars []cel.EnvOption, args map[string]any, outType reflect.Type) (any, error) {
+	env, err := s.env.Extend(vars...)
+	if err != nil {
+		return nil, err
+	}
+	expr = strings.Replace(expr, "$", "__ARG__", -1)
```
shuheiktgw, line 357: nit, `__ARG__ ` appears a few times in the code, so I believe it would be better if we constantize it and explain what it's for

shuheiktgw: nit, might be better if we check the expr contains `__ARG__` before replacing `$` with it, even though that is quite unlikely users use it, but just in case...

generator/templates/server.go.tmpl
```diff
@@ -278,6 +430,30 @@ func recoverError{{ $serviceName }}(v interface{}, rawStack []byte) *{{ $service
 	}
 }
 
+func (s *{{ $serviceName }}) evalCEL(expr string, vars []cel.EnvOption, args map[string]any, outType reflect.Type) (any, error) {
+	env, err := s.env.Extend(vars...)
+	if err != nil {
+		return nil, err
+	}
+	expr = strings.Replace(expr, "$", "__ARG__", -1)
```
shuheiktgw, line 438: nit, might be better if we check the expr contains __ARG__ before replacing $ with it, even though that is quite unlikely users use it...

generator/templates/server.go.tmpl
```diff
@@ -147,6 +158,137 @@ type {{ .Name }} struct {
 
 {{- end }}
 
+// FederationServiceCELTypeHelper
+type FederationServiceCELTypeHelper struct {
+	celRegistry    *celtypes.Registry
+	structFieldMap map[string]map[string]*celtypes.FieldType
+	mapMu          sync.Mutex
```
shuheiktgw, line 165: Instead of using `sync.Mutex`, it may be better to use `sync.RWMutex`. Then we can use `h.mapMu.RLock()` instead of `h.mapMu.Lock()` when we only read from the map for better concurrency.

goccy: Certainly, the method you pointed out seems to good way for current implementation 👍 
However, to improve performance, I'm going to stop using map and replace it with switch-case.

generator/templates/server.go.tmpl
goccy: Thank you. I will verify this in the resolver phase. I also assume that we always evaluate valid expressions in the generator phase ( because generator phase must be evaluated after resolver phase ).

shuheiktgw (approved):
LGTM 🙆‍♂️

PR #32, opened 2023-10-13
Fix a bug with duplicated resolver methods

shuheiktgw:
## NOTE
The problem was if multiple root nodes contained the same messages, [the Resolver#resolveMessageArgument method could insert the same Messages into the same Service multiple times](https://github.com/mercari/grpc-federation/blob/7b0c2fe6328431246a9cc29907e119606132bd4e/resolver/resolver.go#L1728-L1729). As a result, duplicated methods were generated.

Please check the testdata/resolver_overlaps.proto as well to understand how to replicate the problem.


resolver/resolver.go
```diff
@@ -1701,6 +1701,8 @@ func (r *Resolver) resolveMessageArgument(ctx *context, files []*File) {
 	if graph == nil {
 		return
 	}
+
+	serviceMsgMap := make(map[string]map[string]*Message)
```
shuheiktgw, line 1705: Another option may be to change [the Service.Messages and Service.MessageArgs](https://github.com/mercari/grpc-federation/blob/7b0c2fe6328431246a9cc29907e119606132bd4e/resolver/types.go#L40C14-L41) from slice to map, but then sorting it is kind of complicated, so I decided to use the serviceMsgMap instead.

resolver/resolver.go
```diff
@@ -1709,9 +1711,21 @@ func (r *Resolver) resolveMessageArgument(ctx *context, files []*File) {
...
+			if _, exists := serviceMsgMap[svc.FQDN()]; !exists {
+				serviceMsgMap[svc.FQDN()] = make(map[string]*Message)
+			}
+			for _, msg := range r.resolveMessageArgumentRecursive(ctx, root) {
+				serviceMsgMap[svc.FQDN()][msg.FQDN()] = msg
+			}
+		}
+	}
+
+	for fqdn, svc := range r.cachedServiceMap {
```
goccy, line 1725: If you change type of `serviceMsgMap` to `map[*Service]map[string]*Message` or `map[*Service]map[*Message]struct{}` , you can use `serviceMsgMap` here instead of `r.cachedServiceMap` .

Pointer's value to the same service or message are guaranteed to be the same.

goccy (approved):
LGTM !!

PR #37, opened 2023-10-24
Support default for oneof feature

goccy:
- Support `default` case for `grpc.federation.field.oneof` feature.
- Add test cases of validation for `oneof` feature.
- Sort validation errors by line and column number.


validator/validator.go
```diff
@@ -80,12 +81,15 @@ func (v *Validator) Validate(ctx context.Context, file *source.File, opts ...Val
 	if result != nil {
 		outs = v.toValidationOutputByWarnings(dirName, result.Warnings)
 	}
-	if err == nil {
-		return outs
-	}
 	for _, e := range resolver.ExtractIndividualErrors(err) {
 		outs = append(outs, v.toValidationOutputByError(dirName, e))
 	}
+	sort.SliceStable(outs, func(i, j int) bool {
```
shuheiktgw, line 87: nit, it could be one sort operation?

```
if outs[i].Start.Line == outs[j].Start.Line {
  return outs[i].Start.Col < outs[j].Start.Col
}

return outs[i].Start.Line < outs[j].Start.Line
```

goccy: Thank you. I like SliceStable's approach, so I will do this approach if we've performance issues.

shuheiktgw (approved):
LGTM 🙆‍♂️

PR #38, opened 2023-10-24
Add proto for the validation

shuheiktgw:
I've added proto definition for the validation


proto/grpc/federation/federation.proto
```diff
@@ -532,4 +538,57 @@ message MessageFieldValue {
...
+  // Either rule or details must be specified.
+  repeated ValidationErrorDetail details = 4;
+}
+
+// ValidationErrorDetail represents a pair of validation rule and error details to be set.
+message ValidationErrorDetail {
+  // rule specifies validation rule in CEL.
+  string rule = 1;
+  // error_info describes the cause of the error with structured details.
+  repeated google.rpc.ErrorInfo error_info = 3;
```
shuheiktgw, line 575: I'd like to reserve 2 for the messages

compiler/compiler.go
```diff
@@ -113,6 +116,12 @@ func (c *Compiler) Compile(ctx context.Context, file *source.File, opts ...Optio
 					if c.autoImport && strings.HasSuffix(p, grpcFederationFilePath) {
 						return io.NopCloser(bytes.NewBuffer(federation.ProtoFile)), nil
 					}
+					if c.autoImport && strings.HasSuffix(p, googleRPCCodeFilePath) {
```
shuheiktgw, line 119: Now that the plugins do not work without the auto import option enabled, do we need it in the first place...? In other words, are there any cases we need to disable the option?

compiler/compiler.go
```diff
@@ -17,6 +17,7 @@ import (
 	"google.golang.org/protobuf/types/descriptorpb"
 
 	"github.com/mercari/grpc-federation/proto/grpc/federation"
+	"github.com/mercari/grpc-federation/proto_deps/google/rpc"
```
goccy, line 20: I thought the method of creating a `google/rpc` directory under `proto` might also work.
In this case, I thought it would be nice to just specify `-Igithub.com/mercari/grpc-federation/proto` to allow for dependency resolution.

What do you think ?

compiler/compiler.go
goccy: It was possible to disable it if someone do not want implicit dependency resolution.
It may seems good autoImport enabled by default.

buf.gen.yaml
```diff
@@ -2,4 +2,4 @@ version: v1
 plugins:
   - plugin: go
     out: .
-    opt: module=github.com/mercari/grpc-federation
```
shuheiktgw, line 5: We cannot specify the module if the source contains multiple modules, so I updated it to `paths=source_relative`

compiler/compiler.go
shuheiktgw: Makes sense. I've enabled the auto import by default 🙂

.octocov.yml
```diff
@@ -1,11 +1,12 @@
 coverage:
   paths:
     - cover.out
-  acceptable: current >= 60% && diff >= -1.0%
```
shuheiktgw, line 4: I think it is a bit too sensitive so I removed it for now...

compiler/compiler_test.go
```diff
@@ -28,7 +29,11 @@ func TestCompiler(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	const expectedProtoNum = 5 // service.proto, post.proto, user.proto, federation.proto, google/protobuf/descriptor.proto
+	for _, proto := range protos {
+		fmt.Println(*proto.Name)
```
goccy, line 33: Please remove debug codes 🙏

shuheiktgw: 🤦‍♂️

PR #40, opened 2023-10-26
Fix dependency list for oneof feature

goccy:
- Fix import statements and the list of dependency method at the generated codes.
- Run compilation checking at testing of code generator.


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #44, opened 2023-10-30
Support multiple services

goccy:
- Support multiple services
- Support multiple files in same package

When the same message is referenced in multiple files and multiple services, the same message argument structure must be referenced. For this reason, Go's generics are used for the message argument structure so that `DependentClientSet` can be managed on a per-service basis.

- The logic commonly used in the generated code was moved to `github.com/mercari/grpc-federation/grpc/federation` package.
  - Simpler generated code

### Changed File Name

File names will be changed from service name-based to file name-based.

#### BEFORE

- File: `foo.proto`
- Service: `FederationService` and `DebugService`

Generated file
- `federationservice_grpc_federation.go`
- `debugservice_grpc_federation.go`

#### AFTER

- File: `foo.proto`
- Service: `FederationService` and `DebugService`

Generated file
- `foo_grpc_federation.pb.go`

### Changed Resolver and CodeGenerator's I/F

- Outputs the result of Resolver as a `resolver.File` structure
- CodeGenerator's Generate method accepts `resolver.File`


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #47, opened 2023-10-31
Fix oneof

goccy:
Fix problem of generated codes when there was field option and without a message option.


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #49, opened 2023-11-01
Support message validations

shuheiktgw:
## NOTE
This is the very first PR to add a validation feature to the gRPC Federation. The scope is quite limited and the only supported feature is the rule field at the root. If violated, it returns an error.


grpc-federation.yaml
```diff
@@ -0,0 +1,11 @@
+# specify import paths.
```
shuheiktgw, line 1: This is for my local testing so I'm going to remove it later

resolver/graph.go
```diff
@@ -211,33 +211,32 @@ func (g *MessageDependencyGraph) createMessageResolver(ctx *context, node *Messa
 	}
 	ctx.addError(
 		ErrWithLocation(
-			fmt.Sprintf(`%q message has not resolver content`, node.Message.Name),
-			source.MessageLocation(ctx.fileName(), node.Message.Name),
```
shuheiktgw, line 215: The validation nodes do not have a Message reference, so I needed to change it from `node.Message.Name` to `node.BaseMessage.Name`. Considering the context, I believe node.BaseMessage.Name is more appropriate than node.Message.Name since that is where the resolver rule should have been defined but WDYT? 🤔

shuheiktgw: The same with some of the similar changes below

source/location.go
```diff
@@ -907,3 +912,18 @@ func MessageDependencyArgumentInlineLocation(fileName, msgName string, idx, argI
 		},
 	}
 }
+
+// MessageValidationLocation creates location for messages[*].validations[*] in grpc.federation.message.
```
goccy, line 916: `s/messages[*].validations[*]/validations[*]/` ? ( `validations` is top level field of `grpc.federation.message` option )

resolver/resolver_test.go
```diff
@@ -2597,6 +2597,115 @@ func TestOneof(t *testing.T) {
...
+						testutil.NewNameReferenceValueBuilder(
+							ref.Type(t, "org.federation", "Post"),
+							ref.Type(t, "org.federation", "Post"),
+							"post",
+						).Build(t),
+					).Build(t),
+				).
+				SetRule(
+					testutil.NewMessageRuleBuilder().
+						AddMessageDependency(
```
goccy, line 2656: Please add the validation rules by `testutil.MessageRuleBuilder`

resolver/resolver.go
```diff
@@ -2090,6 +2136,25 @@ func (r *Resolver) resolveMessageCELValues(ctx *context, env *cel.Env, msg *Mess
 			}
 		}
 	}
+	for valIdx, validation := range msg.Rule.Validations {
+		if err := r.resolveCELValue(ctx, env, validation.Error.ValidationRule); err != nil {
+			ctx.addError(
+				ErrWithLocation(
+					err.Error(),
+					source.MessageValidationLocation(msg.File.Name, msg.Name, valIdx),
```
goccy, line 2144: The error location to be in the validation rule field, not in the validation itself.

resolver/resolver.go
```diff
@@ -2090,6 +2136,25 @@ func (r *Resolver) resolveMessageCELValues(ctx *context, env *cel.Env, msg *Mess
...
+				),
+			)
+			continue
+		}
+		if validation.Error.ValidationRule.Out.Type != types.Bool {
+			ctx.addError(
+				ErrWithLocation(
+					"validation rule must always return a boolean value",
+					source.MessageValidationLocation(msg.File.Name, msg.Name, valIdx),
```
goccy, line 2153: same as above

resolver/resolver.go
```diff
@@ -986,6 +987,51 @@ func (r *Resolver) resolveMessageAlias(ctx *context, aliasName string) *Message
...
+	e := validation.GetError()
+
+	if e.GetRule() != "" && len(e.GetDetails()) != 0 {
+		return nil, errors.New("cannot set both rule and details at the same time")
+	}
+
+	name := validation.GetName()
+	if name == "" {
+		// the default validation name
+		name = fmt.Sprintf("MessageValidation%d", idx)
```
goccy, line 1018: Use the FQDN of the message to make it unique. e.g.) `fmt.Sprintf("%s_validation_%d", msg.FQDN(), idx)`

resolver/format.go
```diff
@@ -272,6 +276,46 @@ func (d *MessageDependency) protoFormatMessageArgs(opt *ProtoFormatOption) strin
...
+	)
+	return indent + fmt.Sprintf("{\n%s\n%s}", strings.Join(elems, "\n"), indent)
+}
+
+func (v *ValidationRule) protoFormatError(opt *ProtoFormatOption) string {
+	indent := opt.indentFormat()
+	nextOpt := opt.toNextIndentLevel()
+	var elems []string
+	elems = append(elems, nextOpt.indentFormat()+fmt.Sprintf("rule: %q", v.Error.ValidationRule.Expr))
+	return indent + fmt.Sprintf("error: {\n%s\n%s}", strings.Join(elems, "\n"), indent)
```
goccy, line 316: If there is only one element, I want to exclude `:`. e.g.) `error {\n%s\n%s}`

resolver/resolver.go
goccy: Ah, reading the generated code, the FQDN did not seem to be necessary. A name ( please use lower snake case ) that is unique in the message is sufficient.

resolver/graph.go
```diff
@@ -353,6 +367,36 @@ func CreateMessageDependencyGraph(ctx *context, baseMsg *Message) *MessageDepend
 			}
 		}
 	}
+
+	var independentValidationNodes []*MessageDependencyGraphNode
```
goccy, line 371: nits: The name `rootValidationNodes` might be shorter and better

resolver/graph.go
goccy: Thank you ! `BaseMessage` is correct !

goccy (approved):
LGTM !!!

shuheiktgw:
🚀

PR #53, opened 2023-11-08
Fix code generation of log value for message argument

goccy:
Unified code paths for message and message argument.

### DIFF
--- a/generator/code_generator.go
+++ b/generator/code_generator.go
@@ -606,16 +606,7 @@ func (s *Service) setLogValueByMessage(msg *resolver.Message) {
 			Key:   field.Name,
 			Value: s.logValue(field),
 		})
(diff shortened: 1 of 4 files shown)


generator/code_generator.go
```diff
@@ -683,9 +678,7 @@ func (s *Service) setLogValueByMessageArgument(msg *resolver.Message) {
 			Key:   field.Name,
 			Value: s.msgArgumentLogValue(field),
 		})
-		if field.Type.Ref != nil {
```
shuheiktgw, line 686: IIUC, this is the part where some of the log values are missing?

shuheiktgw (approved):
LGTM 🙆‍♂️

generator/code_generator.go
goccy: Yes !

PR #54, opened 2023-11-10
Support error details

shuheiktgw:
I've added support for error details in case of validation failures. Currently, only the precondition failure is supported, but if that looks OK, I'll also implement other detail types. Thanks!


proto/grpc/federation/federation.proto
```diff
@@ -579,23 +579,23 @@ message ValidationErrorDetail {
   // rule specifies validation rule in CEL.
   string rule = 1;
   // error_info describes the cause of the error with structured details.
-  repeated google.rpc.ErrorInfo error_info = 3;
+  repeated google.rpc.ErrorInfo error_infos = 3;
```
shuheiktgw, line 582: If we follow the current naming convention, I believe the field should be named `error_infos` instead of `error_info`, so I renamed it.

resolver/format.go
```diff
@@ -303,24 +303,64 @@ func (v *ValidationRule) ProtoFormat(opt *ProtoFormatOption) string {
...
+	}
+
+	return strings.Join(elems, "\n")
+}
+
+func (v *ValidationErrorDetail) protoFormatDetails(opt *ProtoFormatOption, name string, size int) string {
+	indent := opt.indentFormat()
+	nextOpt := opt.toNextIndentLevel()
+	if size == 1 {
+		return indent + fmt.Sprintf("%s {...}", name)
```
shuheiktgw, line 355: It is a little too time-consuming to support each field for all the types, so I shortcut it for now...

source/location.go
```diff
@@ -173,8 +173,25 @@ type ArgumentOption struct {
...
+	Rule                bool
+	PreconditionFailure *MessageValidationDetailPreconditionFailureOption
+}
+
+type MessageValidationDetailPreconditionFailureOption struct {
+	Idx       int
+	Violation MessageValidationDetailPreconditionFailureViolationOption
+}
+
+type MessageValidationDetailPreconditionFailureViolationOption struct {
```
shuheiktgw, line 192: This is getting a little too much, and we may need some refactoring...

source/location.go
```diff
@@ -173,8 +173,25 @@ type ArgumentOption struct {
...
+	Idx       int
+	FieldName string
```
shuheiktgw, line 194: This is actually slightly different from the other options, but I believe having the field name directly is more concise than creating boolean fields for each field in the PreconditionFailure_Violation in this case. This works fine since all the fields are strings.

resolver/types.go
```diff
@@ -164,8 +164,52 @@ type ValidationRule struct {
...
 type ValidationError struct {
-	Code code.Code
-	Rule *CELValue
+	Code    code.Code
+	Rule    *CELValue
+	Details MessageValidationDetails
+}
+
+// ReferenceNames returns all the unique reference names in the error definition.
+func (v *ValidationError) ReferenceNames() []string {
```
goccy, line 173: Please move this logic to resolver/validation.go, since I want to keep resolver/types.go to type definitions and minimal implementation.

generator/testdata/expected_validation.go
```diff
@@ -9,13 +9,15 @@ import (
 	"runtime/debug"
 	"sync"
 
+	githubproto "github.com/golang/protobuf/proto"
```
goccy, line 12: Is there a reason you created alias ? I would like to use the original packge name as much as possible.

generator/templates/server.go.tmpl
```diff
@@ -777,14 +778,79 @@ if err := eg.Wait(); err != nil {
...
+	{
+	  valueMu.RLock()
+	  _value, err := grpcfed.EvalCEL(s.env, "{{ .Rule }}", envOpts, evalValues, reflect.TypeOf(false))
+	  valueMu.RUnlock()
+	  if err != nil {
+	    return nil, err
+	  }
+	  if !_value.(bool) {
+		  _success = false
+	    {{- range .PreconditionFailures }}
```
goccy, line 804: indent

generator/templates/server.go.tmpl
```diff
@@ -777,14 +778,79 @@ if err := eg.Wait(); err != nil {
...
+	  }
+	  return nil, _stts.Err()
+	}
+	{{- end }}
+}
+{{- end }}
+
+{{- define "preconditionFailure" }}
+{
+  var _validations []*errdetails.PreconditionFailure_Violation
```
goccy, line 824: indent

generator/templates/server.go.tmpl
```diff
@@ -777,14 +778,79 @@ if err := eg.Wait(); err != nil {
...
+	    return nil, err
+	  }
+	  _validations = append(_validations, &errdetails.PreconditionFailure_Violation{
+	    Type: _type.(string),
+	    Subject: _subject.(string),
+	    Description: _description.(string),
+	  })
+	}
+	{{- end }}
+  _details = append(_details, &errdetails.PreconditionFailure{
```
goccy, line 852: indent

generator/templates/server.go.tmpl
```diff
@@ -777,14 +778,79 @@ if err := eg.Wait(); err != nil {
 
 {{- define "messageValidation" }}
 {
+  {{- if .Error.HasRule }}
```
goccy, line 781: indent

goccy (commented):
It does not have to do this PR, but please add a test to the validator packge.

_examples/12_simple_validation/proto/federation/federation.proto
```diff
@@ -39,7 +33,23 @@ message GetPostResponse {
       {
         error: {
           code: FAILED_PRECONDITION,
-          rule: "post.title == 'some-title'",
+          details: {
+            rule: "$.id == 'correct-id'",
+            precondition_failures: {
+              violations: [
+                {
+                  type: "'type1'"
```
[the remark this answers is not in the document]
shuheiktgw: It would indeed be more user-friendly if they didn't have to wrap it with `''`. How about we first parse it as CEL; if that returns an error, we treat the field as a string. We may add a new field to the CELValue to indicate the field could be either CEL or plane string.

generator/testdata/expected_validation.go
shuheiktgw: That actually conflicts with ["google.golang.org/protobuf/proto"](https://github.com/mercari/grpc-federation/blob/7ec704f932af243ba27a311bac37c1d47a6900f7/generator/templates/server.go.tmpl#L17)...

generator/templates/server.go.tmpl
shuheiktgw: 🙇

generator/testdata/expected_validation.go
goccy: naruhodo. ( Message V1 and V2 problem...)
Oh, `google.golang.org/protobuf/proto` does not seem to be used, so it may be deleted.

generator/templates/server.go.tmpl
```diff
@@ -776,15 +777,80 @@ if err := eg.Wait(); err != nil {
...
+			return nil, err
+		}
+		if !_value.(bool) {
+			_success = false
+			{{- range .PreconditionFailures }}
+			{{- template "preconditionFailure" . }}
+			{{- end }}
+		}
+	}
+  {{- end }}
```
goccy, line 809: indent

shuheiktgw: 🙇

shuheiktgw: kondokoso...

goccy (approved):

shuheiktgw:
🚀

PR #57, opened 2023-11-17
Supports variable definition

goccy:


resolver/resolver.go
```diff
@@ -992,8 +1024,245 @@ func (r *Resolver) resolveMessageRule(ctx *context, msg *Message, ruleDef *feder
...
+		Expr:     r.resolveMapIteratorExpr(ctx, def),
+	}
+}
+
+func (r *Resolver) resolveMapIterator(ctx *context, def *federation.Iterator) *Iterator {
+	name := def.GetName()
+	if name == "" {
+		ctx.addError(
+			ErrWithLocation(
+				"map iterator name must be speficied",
```
shuheiktgw, line 1096: nit, speficied -> specified. The same with the one below

resolver/resolver.go
```diff
@@ -2294,6 +2579,29 @@ func (r *Resolver) resolveMessageCELValues(ctx *context, env *cel.Env, msg *Mess
 			r.resolveMessageValidationErrorDetailCELValues(ctx, env, msg, valIdx, detIdx, detail)
 		}
 	}
+	for idx, varDef := range msg.Rule.VariableDefinitions {
+		if varDef.Expr == nil {
```
shuheiktgw, line 2583: Is ignoring `varDef.If` intentional? I know the feature is not supposed to be supported in this PR but it seems like the other CELValue fields are resolved here, I just wanted to make sure...

resolver/resolver.go
```diff
@@ -2415,6 +2723,181 @@ func (r *Resolver) resolveMessageCELValues(ctx *context, env *cel.Env, msg *Mess
...
+			)
+		}
+		expr.Type = expr.By.Out
+	case expr.Message != nil:
+		for argIdx, arg := range expr.Message.Args {
+			if arg.Value == nil {
+				continue
+			}
+			if err := r.resolveCELValue(ctx, env, arg.Value.CEL); err != nil {
+				if arg.Value.CEL != nil && arg.Value.Inline {
```
shuheiktgw, line 2870: nit, IIUC if `arg.Value.CEL` is nil, `resolveCELValue` never returns an error, so `arg.Value.CEL != nil` is unnecessary.

shuheiktgw (approved):
I left some minor comments, but overall it looks great to me! 👍

PR #59, opened 2023-11-20
Fix how to get extension rule

goccy:
- Fix a bug that caused failure to retrieve extensions when multiple extensions were set.
- Refactor how to get extension rule by generics


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #60, opened 2023-11-20
Fix duplicate cast functions are generated

goccy:
If cast function generated for each message, it is duplicated. So it needs to make unique,  generate for each service.


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #61, opened 2023-11-20
Fix generator

goccy:
- Fix proto file relative path
- Add the file that does not have service to generate target
- Use resolver.File instance to call `toTypeText` instead of resolver.Service ( refactor )


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #69, opened 2023-11-29
Support map operation for message

goccy:
- Add map operator's test case
- Add MapResolver type to code generator to generate map operation
- Refactor MessageExpr type processing


generator/templates/server.go.tmpl
```diff
@@ -653,7 +653,70 @@ return grpcfed.WithRetry[{{ .Resolver.ReturnType }}](b, func() (*{{ .Resolver.Re
...
+value, err := grpcfed.EvalCEL(s.env, "{{ $by.Expr }}", envOpts, evalValues, reflect.TypeOf({{ .Resolver.ZeroValue }}))
+valueMu.RUnlock()
+return value, err
+{{- else if .Resolver.IsMap }}
+{{- $mapResolver := .Resolver.MapResolver }}
+env, err := s.env.Extend(cel.Variable("{{ $mapResolver.IteratorName }}", {{ $mapResolver.IteratorCELType }}))
+if err != nil {
+	return nil, err
+}
+valueMu.RLock()
```
shuheiktgw, line 666: IIUC, the lock won't be unlocked if it returns an error at L689?

generator/templates/server.go.tmpl
```diff
@@ -653,7 +653,70 @@ return grpcfed.WithRetry[{{ .Resolver.ReturnType }}](b, func() (*{{ .Resolver.Re
...
+		{{- end }}
+		{{- else }}
+		args.{{ .Name }} = {{ .Value }}
+		{{- end }}
+	}
+	{{- end }}
+	{{- end }}
+	iterValue, err := s.{{ $mapResolver.Caller }}({{ .Ctx }}, args)
+	if err != nil {
+		return nil, err
```
shuheiktgw, line 713: I guess we need to have `grpcfed.RecordErrorToSpan(ctx, err)` here as well?

PR #70, opened 2023-11-30
Remove unnecessary reference name checks

shuheiktgw:
SSIA. Some reference name checks were unnecessary and caused the complication to fail when we referred to the enum values from CEL. So we decided to remove them since CEL checks their existence anyway.

### DIFF
--- a/validator/validator_test.go
+++ b/validator/validator_test.go
@@ -273,6 +273,7 @@ testdata/missing_enum_value_alias.proto:56:3: specified "alias" in grpc.federati
 56:    POST_TYPE_BAR = 2;
        ^
 `},
+		{file: "valid_enum_value_reference.proto", expected: ""},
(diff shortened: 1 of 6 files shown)


validator/validator_test.go
```diff
@@ -486,8 +487,15 @@ testdata/invalid_validation_localized_message.proto:54:26: message must always r
 			if err != nil {
 				t.Fatal(err)
 			}
-			actual := "\n" + validator.Format(v.Validate(ctx, file))
-			if diff := cmp.Diff(actual, test.expected); diff != "" {
+			got := validator.Format(v.Validate(ctx, file))
```
shuheiktgw, line 490: I want to use the same test table to test if we don't receive any errors, so I've slightly updated it...

goccy (approved):
Thank you for fixing ! LGTM 👍

shuheiktgw:
🚀

PR #72, opened 2023-12-01
Rename validation field from rule to if

shuheiktgw:
SSIA. Renamed it from rule to if for usability.


proto/grpc/federation/federation.proto
```diff
@@ -627,8 +628,8 @@ message ValidationError {
 
 // ValidationErrorDetail represents a pair of validation rule and error details to be set.
 message ValidationErrorDetail {
-  // rule specifies validation rule in CEL.
-  string rule = 1;
+  // if specifies validation rule in CEL. If the condition is true, the validation returns an error.
```
goccy, line 631: 👍

goccy (approved):

shuheiktgw:
🚀

PR #74, opened 2023-12-01
Fix validation with custom resolver

shuheiktgw:
There were several problems with the validation, so I fixed them.

1. If we define a custom resolver with the validation, it tries assigning a variable to a non-existing field
2. Validation does not support the message field
3. Validation does not support the if statement
4. Validation does not use the variable `sg`, it could raise an unused variable error


generator/code_generator.go
```diff
@@ -1130,13 +1130,13 @@ func (m *Message) DeclVariables() []*DeclVariable {
 	}
 	for _, group := range m.Message.MessageResolvers() {
 		for _, r := range group.Resolvers() {
-			if r.VariableDefinition != nil {
-				if !r.VariableDefinition.Used {
+			if varDef := r.VariableDefinition; varDef != nil {
```
shuheiktgw, line 1133: The linter keeps complaining me about it so I fixed it

resolver/resolver.go
```diff
@@ -2547,6 +2548,8 @@ func (r *Resolver) resolveVariableExprCELValues(ctx *context, env *cel.Env, expr
 		for detIdx, detail := range expr.Validation.Error.Details {
 			r.resolveMessageValidationErrorDetailCELValues(ctx, env, ctx.msg, ctx.defIndex(), detIdx, detail)
 		}
+		// This is a dummy type since the output from the validation is not supposed to be used (at least for now)
+		expr.Type = BoolType
```
shuheiktgw, line 2552: This is a bit hacky, but I needed to set some type for MessageResolver#ZeroValue, which is called [here](https://github.com/mercari/grpc-federation/blob/ffeb8a3c7e94ab319b2fa9de363aa7db2f4a5a0f/generator/templates/server.go.tmpl#L617)

generator/templates/server.go.tmpl
```diff
@@ -777,9 +780,6 @@ return nil, err
 {{ .Resolver.ProtoComment }}
 */
 {
-	{{- if .Resolver.IsValidation }}
```
shuheiktgw, line 780: I've moved it into the `callResolver` template

1. To reuse the code for the `if` statement
2. To use the `sg` variable from validations

generator/code_generator.go
```diff
@@ -1479,17 +1479,19 @@ func (m *Message) CustomResolverArguments() []*Argument {
...
-				args = append(args, &Argument{
-					Name:  util.ToPublicGoVariable(name),
-					Value: toUserDefinedVariable(name),
-				})
-				argNameMap[name] = struct{}{}
+			varDef := msgResolver.VariableDefinition
+			if varDef == nil || !varDef.Used {
+				continue
 			}
+			name := msgResolver.VariableDefinition.Name
```
goccy, line 1486: nits: varDef.Name ?

resolver/resolver.go
goccy: This type should be error type, but it is ok now

goccy (approved):

PR #75, opened 2023-12-04
Support list library

goccy:
The list library adds the following two macros.

- `reduce` : reduce operation for list value
  - e.g.) `[1, 2, 3, 4, 5].reduce(accum, cur, accum + cur, 0)` => `15`
  - FYI: https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

- `first` : return first matched value from list value
  - e.g.) `[1, 2, 3, 4, 5].first(v, v % 2 == 0)` => `2`

### DIFF
--- a/grpc/federation/cel.go
+++ b/grpc/federation/cel.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 
 	"github.com/google/cel-go/cel"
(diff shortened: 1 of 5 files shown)


shuheiktgw (approved):
🙆‍♂️

PR #76, opened 2023-12-05
Support uuid and rand library for CEL operation

goccy:
When sending a gRPC request to a dependent service, the idempotency-key may be sent. In this case, uuid is often used, so it is supported as a standard library.

- `grpc.federation.uuid.new().string()` : generate uuid and returns it as a string type
- `grpc.federation.uuid.newRandomFromRand(grpc.federation.rand.new(grpc.federation.rand.newSource(1)))` : generates uuid from fixed seed value


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #79, opened 2023-12-08
Fix custom resolver for map definition

goccy:
Fix bug that custom resolver list was incomplete when map or validation was defined


shuheiktgw (approved):
LGTM! 🙆‍♂️

PR #80, opened 2023-12-11
Fix suffix of ctx variable

goccy:
We use `ContextEval` API to eval CEL to escape `unused variable "ctx"` error.
( because If only `grpcfed.EvalCEL` is used, then `ctx` is no longer needed )


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #81, opened 2023-12-11
Fix type conversion for Standard CEL library types

goccy:
- Add test case / examples for fixing type conversion error for standard cel library types
- Change interface for `grpc.federation.time.UTC` and `grpc.federation.time.LOCAL` from constant variables to function  because I got `unsupported constant kind` error ( FYI: https://github.com/google/cel-go/blob/v0.18.1/common/ast/conversion.go#L631 )


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #84, opened 2023-12-27
Update docs

goccy:
Currently, gRPC Federation status is **Beta**.
So we should remove document for Alpha.


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #89, opened 2024-01-18
Add gRPC metadata library

shuheiktgw:
Add the metadata library to interact with gRPC metadata from the CEL expression.


grpc/federation/cel/metadata.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func createMetadataID(name string) string {
+	return createID(MetadataPackageName, name)
+}
+
+func (lib *MetadataLibrary) CompileOptions() []cel.EnvOption {
+	opts := []cel.EnvOption{
+		cel.Function(
+			createMetadata("INCOMING"),
```
goccy, line 39: I’d like to use UPPER CASE name for constant value. So in this case, I think better to use lower case name ( `incoming` ).

shuheiktgw:
🚀

PR #90, opened 2024-01-22
Add the header field to CallExpr

shuheiktgw:
SSIA. I've added a field called header for the header configuration.


proto/grpc/federation/federation.proto
```diff
@@ -179,6 +179,8 @@ message CallExpr {
   optional string timeout = 3;
   // retry specifies the retry policy if the method call fails.
   optional RetryPolicy retry = 4;
+  // header configures the response header metadata.
+  repeated HeaderConfig header = 5;
```
shuheiktgw, line 183: Should I name it `headers`? I remember we talked about it before and went with a singular field name at that time.

proto/grpc/federation/federation.proto
```diff
@@ -216,6 +218,16 @@ message RetryPolicyExponential {
...
 }
 
+// HeaderConfig define the response header metadata configuration.
+message HeaderConfig {
+  // condition to set the header. If true, the header is set. The condition is evaluated as CEL expression.
+  optional string if = 1;
+  // a kye for the metadata.
+  string key = 2;
+  // a value for the metadata. The value is evaluated as CEL expression.
+  repeated string value = 3;
```
shuheiktgw, line 228: The same here 
https://github.com/mercari/grpc-federation/pull/90/files#r1461410777

proto/grpc/federation/federation.proto
goccy: `header` is good 👍

goccy (approved):
LGTM !

shuheiktgw:
We decided to postpone adding the API. Closing.

PR #92, opened 2024-01-23
Enable the ContextualLibrary

shuheiktgw:
Enabled the ContextualLibrary I've added in https://github.com/mercari/grpc-federation/pull/89.


grpc/federation/cel/lib.go
```diff
@@ -35,7 +38,7 @@ func IsStandardLibraryType(typeName string) bool {
 }
 
 func (lib *Library) LibraryName() string {
-	return "grpc.federation"
+	return lib.name
```
shuheiktgw, line 41: I've noticed that we cannot use the same name between the static and contextual libraries since the libraries are singleton and immutable.

https://github.com/google/cel-go/blob/df072086206afaba0112df1facdd7cf7183af8a3/cel/library.go#L76-L81

I've circumvented the problem by using the different names between the two. Another option may be to stop using the SingletonLibrary interface, but what do you think? 🤔

goccy: The library name does not affect the gRPC Federation's user, so I think this is good 👍

goccy (approved):

PR #93, opened 2024-01-25
Fix imports

goccy:
Could not add to Buf Plugin because `golang.org/x/tools/imports` depend on `go` commands. So, we can determine which libraries to import by ourselves.

Context: https://github.com/bufbuild/plugins/issues/991#issuecomment-1908792464


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #97, opened 2024-02-02
Fix extension number

goccy:
gRPC Federation has been registered in the Protobuf Registry 🎉 
Number `1187` has been assigned.

- Registry: https://github.com/protocolbuffers/protobuf/blob/main/docs/options.md
- Request PR: https://github.com/protocolbuffers/protobuf/pull/15644


shuheiktgw (approved):
🎉

PR #99, opened 2024-02-06
Support serialization of resolver.File

goccy:
We will be able to serialize the `resolver.File` to use the wasm plugin in the CodeGeneration Pipeline.
I will create the plugin system with the next PR.

- `generator.proto` defines schema to serialize `resolver.File`
- `grpc/federation/generator/plugin/generator.pb.go` is a generated file by `generator.proto`
- `grpc/federation/generator/{encode.go,decode.go}` are helper files to serialize `resolver.File`
  - `resolver.File` has some cyclic pointer references, so it can't serialize as is


shuheiktgw (approved):
LGTM 🙆‍♂️

goccy:
Thank you for reviewing !

PR #105, opened 2024-02-19
Improve semantic highlighting

goccy:
Syntax highlighting is now available in quoted value. This is especially effective in CEL value.

semantic_highlighting


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #109, opened 2024-02-22
Resolve GRPCError for CallExpr

goccy:
- Add an error handler implementation to the Resolver. ( The generator will be implemented later )
- When resolving, the `LocationBuilder` instance passes to the almost method


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #111, opened 2024-02-28
Add error variable

goccy:
Add the `error` variable that can be used in error block.


validator/validator_test.go
```diff
@@ -501,6 +517,7 @@ testdata/invalid_validation_localized_message.proto:54:26: message must always r
 			}
 
 			if diff := cmp.Diff("\n"+got, test.expected); diff != "" {
+				fmt.Println(got)
```
shuheiktgw, line 520: Is this intentional, or is it just a leftover from debugging? 🤔

shuheiktgw (approved):
LGTM 🙆‍♂️

validator/validator_test.go
goccy: I forgot to remove it !! Thanks 🙏

PR #112, opened 2024-02-28
Support name and value accessor for enum value

goccy:
- `<enum-fqdn>.name(<enum-value-fqdn>)`: get enum name from value
- `<enum-fqdn>.value('<enum-value-name>')`: get enum value from name


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #113, opened 2024-02-29
Improve `Goto Definition` feature for language server

goccy:
- Makes more targets to jump to definition source
- Use `protocol.LocationLink`


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #123, opened 2024-03-09
chore: upgrade protocompile, buf, bufcli

evacchi:
Update these dependencies to their latest version and re-generate the sources to prepare for the upcoming wazero release. Will follow-up with a PR to upgrade wazero separately.


goccy:
Thank you for your contribution !!! LGTM 👍

PR #125, opened 2024-03-12
Fix import errors

shuheiktgw:
I've tackled the following two problems:

1. `context` and `grpcfed` libraries are imported unnecessarily even if a proto file does not contain any service or plugin definitions.
2.  `generateByProtogenGoGRPC` panics when running against a file that does not contain any service definitions.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x1013506b4]

goroutine 1 [running]:
google.golang.org/protobuf/compiler/protogen.(*GeneratedFile).Content(0x0)
	/Users/shuheiktgw/go/1.22.0/pkg/mod/google.golang.org/protobuf@v1.32.0/compiler/protogen/protogen.go:1045 +0x24
github.com/mercari/grpc-federation/generator.(*Generator).generateByProtogenGoGRPC(0x1400033c000, 0x14001011170)
	/Users/shuheiktgw/go/1.22.0/pkg/mod/github.com/mercari/grpc-federation@v0.13.1/generator/generator.go:563 +0x16c
github.com/mercari/grpc-federation/generator.(*Generator).generateByPlugin(0x1400033c000, {0x101b14ac0?, 0x1025340e0?}, 0x14001011170, 0x1400035c5a0)
	/Users/shuheiktgw/go/1.22.0/pkg/mod/github.com/mercari/grpc-federation@v0.13.1/generator/generator.go:267 +0x15c
```

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -560,6 +560,9 @@ func (g *Generator) generateByProtogenGoGRPC(r *PluginRequest) (*pluginpb.CodeGe
 			continue
 		}
 		generatedFile := runProtogenGoGRPC(r.genplugin, f, true)
+		if generatedFile == nil {
+			continue
+		}
 		content, err := generatedFile.Content()
 		if err != nil {
 			return nil, err
--- a/generator/code_generator.go
+++ b/generator/code_generator.go
@@ -320,7 +320,7 @@ func (f *File) StandardImports() []*Import {
 	}
 
 	pkgs := []*Import{
-		{Path: "context", Used: true},
+		{Path: "context", Used: existsServiceDef || existsPluginDef},
 		{Path: "encoding/json", Used: existsPluginDef},
 		{Path: "io", Used: existsServiceDef},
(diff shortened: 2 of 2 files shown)


generator/generator.go
```diff
@@ -560,6 +560,9 @@ func (g *Generator) generateByProtogenGoGRPC(r *PluginRequest) (*pluginpb.CodeGe
 			continue
 		}
 		generatedFile := runProtogenGoGRPC(r.genplugin, f, true)
+		if generatedFile == nil {
```
shuheiktgw, line 563: runProtogenGoGRPC [could return nil](https://github.com/mercari/grpc-federation/blob/9d1cb726183b8af1779af95d10d46d869ac22261/generator/protoc_gen_go_grpc.go#L136-L138) if a file does not contain any services.

goccy (approved):

PR #129, opened 2024-03-14
Use f.File.Enums instead of f.enums for import

shuheiktgw:
We should have used `f.File.Enums` instead of `f.enums` since `f.enums` contain all the nums defined in the package.


shuheiktgw:
Ah this is wrong

PR #130, opened 2024-03-14
Remove enums from import for non-service files

shuheiktgw:
f.enums contain all the enums defined in the package, and Enums are used only from a File contains Services, so we need to exclude enums from import when a file does not contain any services.


goccy (approved):

Makefile
```diff
@@ -75,7 +75,7 @@ fmt/buf:
 	buf format --write
 
 .PHONY: generate
-generate: generate/buf generate/examples
+generate: build generate/buf generate/examples
```
shuheiktgw, line 78: I want the build task to run it automatically before running the generate task 🙏

goccy (approved):

PR #134, opened 2024-03-18
Fix nested message handling

goccy:
Fixed a problem where nested messages could not be referenced by "message" expressions


shuheiktgw (approved):
🙆‍♂️

PR #135, opened 2024-03-19
Add buf/lint

shuheiktgw:
SSIA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ tools:
 		github.com/golangci/golangci-lint/cmd/golangci-lint
 
 .PHONY: lint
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -68,6 +68,15 @@ lint/gomod: tidy
 		exit 1; \
 	fi
 
+lint/buf: fmt/buf
```
shuheiktgw, line 71: Simply running `buf lint` raises too many errors and is not usable. So I'd like to run buf fmt and check if any diff exists instead at least for now.

goccy (approved):

shuheiktgw:
🙏

PR #139, opened 2024-04-03
Fix location when a message does not exist

shuheiktgw:
The location information was not correct when a user specified non-existing messages, so I fixed it.


goccy (approved):
LGTM !!!

shuheiktgw:
🙏

PR #140, opened 2024-04-04
Remove service dependencies

goccy:
## Why

When implementing the WebAssembly Plugin, the presence of the gRPC client in message argument will cause it to become a blocker. Also, I don't want to call gRPC method in custom resolver.

## What

- Remove process of resolving service dependencies ( `(grpc.federation.service).dependencies` )
- Stop passing gRPC client to custom resolver


resolver/resolver.go
```diff
@@ -1116,35 +1057,11 @@ func (r *Resolver) resolveEnumRules(ctx *context, enums []*Enum) {
 	}
 }
 
-func (r *Resolver) resolveServiceRule(ctx *context, def *federation.ServiceRule, builder *source.ServiceOptionBuilder) *ServiceRule {
+func (r *Resolver) resolveServiceRule(_ *context, def *federation.ServiceRule, _ *source.ServiceOptionBuilder) *ServiceRule {
```
shuheiktgw, line 1060: Are there any reasons you want to keep those unused arguments?

_examples/16_code_gen_plugin/Makefile
```diff
@@ -39,4 +39,4 @@ jaeger/stop:
 	@docker stop $(shell docker ps -q  --filter ancestor=$(JAEGER_IMAGE))
 
 build/wasm:
-	tinygo build -o plugin.wasm -target wasi -no-debug -scheduler=none ./cmd/plugin
+	GOOS=wasip1 GOARCH=wasm go build -o plugin.wasm ./cmd/plugin
```
shuheiktgw, line 42: 👀

shuheiktgw (approved):
LGTM 🙆‍♂️

PR #141, opened 2024-04-05
Fix enum type conversion

goccy:
- Evaluate return type of `<enum>.value(<enum-value>)` CEL function as `opaque<enum>` type instead of `int` type
- Generate type conversion between `enum` and `int` or `uint`


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #142, opened 2024-04-05
Add validation

goccy:
## What

- validate message argument type for multiple calls
- validate alias message
- validate field type conversion

As we cannot use `by` in an aliased message, I fixed test and example.


shuheiktgw (approved):
🙆‍♂️

PR #143, opened 2024-04-08
Skip if the candidates' file name matches unexpectedly

shuheiktgw:
The comment and the test explain it all, but Resolver#Candidates could panic if the file name suffix matches unexpectedly since `r.cachedMessageMap[fmt.Sprintf("%s.%s", protoPkgName, loc.Message.Name)]` returns nil in that case.


goccy (approved):

shuheiktgw:
:ty:

PR #144, opened 2024-04-09
Support JetBrains plugin

shuheiktgw:
SSIA. I've added the JetBrain plugin support using LSP.


.github/workflows/release-jetbrains.yml
```diff
@@ -0,0 +1,80 @@
+name: release-jetbrains
+on:
+  push:
+    tags:
+      - "jetbrains/v*"
```
shuheiktgw, line 5: I'd like to kick the pipeline when a release with jetbrains/vxxx tag is created.

.github/workflows/release-jetbrains.yml
```diff
@@ -0,0 +1,80 @@
+name: release-jetbrains
```
shuheiktgw, line 1: The workflow originally came from https://github.com/JetBrains/intellij-platform-plugin-template/blob/main/.github/workflows/release.yml

.github/workflows/release-jetbrains.yml
```diff
@@ -0,0 +1,80 @@
...
+        run: ./gradlew publishPlugin
+
+      # Upload artifact as a release asset
+      - name: Upload Release Asset
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh release upload ${{ github.event.release.tag_name }} ./build/distributions/*
+
+      # Create a pull request
+      - name: Create Pull Request
```
goccy, line 55: Why create a pull request ?

shuheiktgw: The workflow [updates the changelog](https://github.com/JetBrains/intellij-platform-plugin-template/blob/9726f523c5a41e275fc1fd9f52e01a885ad8e93f/.github/workflows/release.yml#L45-L60) based on the release body, so it needs to update the original changelog. Even though it is nice to have the changelog, I think we don't really need it since it will contain many unrelated changes. Let me remove it 🙆‍♂️

goccy (approved):
LGTM !

shuheiktgw:
🚀

PR #145, opened 2024-04-09
Re-architect wasm plugin system ( remove dependency on TinyGo )

goccy:
Remove dependency on TinyGo by creating WebAssembly using Go's WASI Preview1.


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #146, opened 2024-04-10
Fix error locations

shuheiktgw:
There were several cases where the error locations were wrong, so I fixed them.

- Nested message alias option
- Nested message option
- Nested message field option
- Nested message cycle detection

### DIFF
--- a/resolver/resolver.go
+++ b/resolver/resolver.go
@@ -734,7 +734,9 @@ func (r *Resolver) resolveRule(ctx *context, files []*File) {
 	for _, file := range files {
 		ctx := ctx.withFile(file)
 		r.resolveServiceRules(ctx, file.Services)
-		r.resolveMessageRules(ctx, file.Messages)
+		r.resolveMessageRules(ctx, file.Messages, func(name string) *source.MessageBuilder {
+			return source.NewMessageBuilder(file.Name, name)
+		})
 		r.resolveEnumRules(ctx, file.Enums)
 	}
 }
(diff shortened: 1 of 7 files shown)


resolver/resolver.go
```diff
@@ -3517,6 +3522,14 @@ func (r *Resolver) trimPackage(pkg *Package, name string) string {
 	return strings.TrimPrefix(name, fmt.Sprintf("%s.", pkg.Name))
 }
 
+func newMessageBuilderFromMessage(message *Message) *source.MessageBuilder {
```
shuheiktgw, line 3525: In some cases, specifically when traversing the dependency graph and all the messages, the resolver does not visit the messages based on the proto definition, so we must build the location builder from the Message...

resolver/resolver.go
```diff
@@ -913,10 +917,10 @@ func (r *Resolver) validateName(name string) error {
...
 	for _, msg := range msgs {
 		ctx := ctx.withFile(msg.File).withMessage(msg)
-		builder := source.NewMessageBuilder(msg.File.Name, msg.Name)
-		r.validateMessageAlias(ctx, msg, builder)
-		r.validateMessageFields(ctx, msg, builder)
-		r.validateMessages(ctx, msg.NestedMessages)
+		mb := newMessageBuilderFromMessage(msg)
+		r.validateMessageAlias(ctx, msg, mb)
+		r.validateMessageFields(ctx, msg, mb)
+		// Don't have to check msg.NestedMessages since r.allMessages(files) already includes nested messages
```
shuheiktgw, line 923: We used to validate the nested messages multiple times

goccy (approved):
LGTM !!!! 👍

PR #147, opened 2024-04-10
Improve performance ( add cache mechanism for CEL compilation )

goccy:
- In `grpcfed.EvalCEL` , we found that `cel-go`'s `(*Env).Extend` was taking a long time. Therefore, we cache and reuse AST for CEL and uses `cel.NewCustomEnv` function instead of `(*Env).Extend` .
- Reviewed the lifecycle of the wasm plugin ( Thread safe / Memory safe )
- Add example 18 to test performance


resolver/resolver.go
```diff
@@ -2879,6 +2902,24 @@ func (r *Resolver) resolveCELValue(ctx *context, env *cel.Env, value *CELValue)
 	if err != nil {
 		return err
 	}
+	var useContextLib bool
+	for _, ref := range checkedExpr.GetReferenceMap() {
+		for _, overloadID := range ref.GetOverloadId() {
+			for _, prefix := range r.ctxOverloadIDPrefixes {
+				if strings.HasPrefix(overloadID, prefix) {
+					useContextLib = true
```
shuheiktgw, line 2910: nit, we can break this triple for loop here since we don't have to check the rest of the refs

shuheiktgw (approved):
LGTM!

PR #154, opened 2024-04-17
Support `if` field for `retry`

goccy:
Add `if` spec to control the retry process.


source/file.go
```diff
@@ -1222,6 +1170,12 @@ func (f *File) nodeInfoByRetry(node *ast.MessageLiteralNode, retry *RetryOption)
 	for _, elem := range node.Elements {
 		fieldName := elem.Name.Name.AsIdentifier()
 		switch {
+		case retry.If && fieldName == "if":
+			value, ok := elem.Val.(*ast.StringLiteralNode)
+			if !ok {
+				return nil
+			}
+			return f.nodeInfo(value)
```
shuheiktgw, line 1178: nit, it would be better if we have a simple validation test to make sure this works as expected.

shuheiktgw (approved):
Left some minor comments, but LGTM 🙆‍♂️

PR #161, opened 2024-04-22
Support `ignore_and_response` for error handling

goccy:
`ignore_and_response` ignore the error if the condition in the `if` field is `true` and it returns response specified in CEL.
The evaluation value of CEL must always be the same as the response message type.
`ignore` and `ignore_and_response` cannot be specified same.


proto/grpc/federation/federation.proto
```diff
@@ -203,7 +203,14 @@ message GRPCError {
...
   repeated GRPCErrorDetail details = 5;
   // ignore ignore the error if the condition in the "if" field is true and "ignore" field is set to true.
+  // When an error is ignored, the returned response is always null value.
+  // If you want to return a response that is not null, please use `ignore_and_response` feature.
+  // Therefore, `ignore` and `ignore_and_response` cannot be specified same.
   optional bool ignore = 6;
+  // ignore_and_response ignore the error if the condition in the "if" field is true and it returns response specified in CEL.
+  // The evaluation value of CEL must always be the same as the response message type.
+  // `ignore` and `ignore_and_response` cannot be specified same.
+  optional string ignore_and_response = 7;
```
shuheiktgw, line 213: IMHO, what we really want to express here is constructing arbitrary messages, and `ignore: true` is one of the special cases where the message is null. So we might not need the two fields (`ignore` and `ignore_and_response`) and merge them into one like `response`.

resolver/resolver.go
```diff
@@ -1415,21 +1415,44 @@ func (r *Resolver) resolveValidationExpr(ctx *context, def *federation.Validatio
...
+	if res := def.GetIgnoreAndResponse(); res != "" {
+		if def.GetIgnore() {
+			ctx.addError(
+				ErrWithLocation(
+					`cannot set both "ignore" and "ignore_and_response"`,
+					builder.WithIgnore().Location(),
+				),
+			)
+		}
+		if def.GetIgnore() {
```
shuheiktgw, line 1434: nit, this if statement can move to the `if def.GetIgnore()` block above

shuheiktgw:
Do we need to update encode.go and decode.go?
- https://github.com/mercari/grpc-federation/blob/faec6a8e107603dc1177b3256c2293a63dcd01aa/grpc/federation/generator/encode.go#L1094
- https://github.com/mercari/grpc-federation/blob/faec6a8e107603dc1177b3256c2293a63dcd01aa/grpc/federation/generator/decode.go#L1104

proto/grpc/federation/federation.proto
goccy: I see, but in my opinion, `ignore: true` is sufficient for most use cases. I think it is more of a special case to return a non-null message when ignoring an error.
Therefore, we need to consider whether we should always enforce a specification such as `"null"` .
I also believe that the name `response` should be used with caution because it has a different meaning when used in an `error` block ( It could be misinterpreted as an error value. ).

shuheiktgw (approved):
👍

PR #163, opened 2024-04-25
Support versioning system for CEL plugin

goccy:
- Add a protocol version for validating cel plugin at startup.
- Changes wasm plugin configuration interface to accept io.Reader.
- Add versioning marker to the generated file


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #165, opened 2024-05-03
Fix cel plugin

goccy:
- Add error handling for wazero's instantiate method
- Make the constructor of generator.File public

### DIFF
--- a/grpc/federation/cel/plugin.go
+++ b/grpc/federation/cel/plugin.go
@@ -113,25 +113,33 @@ func (p *CELPlugin) CreateInstance(ctx context.Context, celRegistry *types.Regis
 	modCfg := wazero.NewModuleConfig().
 		WithStdin(stdinR).
 		WithStdout(stdoutW).
-		WithStderr(os.Stderr)
+		WithStderr(os.Stderr).
+		WithArgs("plugin")
 
-	go p.wasmRuntime.InstantiateModule(ctx, p.mod, modCfg) //nolint: errcheck
+	instanceModErrCh := make(chan error)
+	go func() {
+		_, err := p.wasmRuntime.InstantiateModule(ctx, p.mod, modCfg)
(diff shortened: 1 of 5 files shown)


grpc/federation/cel/plugin.go
```diff
@@ -184,8 +192,27 @@ func (i *CELPluginInstance) ValidatePlugin(ctx context.Context) error {
...
+		return errors.New("grpc-federation: plugin has already been closed")
+	}
+
+	writeCh := make(chan error)
+	go func() {
+		_, err := i.stdin.Write(cmd)
+		writeCh <- err
+	}()
+	select {
+	case err := <-i.instanceModErrCh:
```
shuheiktgw, line 206: What if CELPluginInstance#write is called before `p.wasmRuntime.InstantiateModule` finishes running? 🤔 Should we somehow wait until the instantiation is completed...?

shuheiktgw: Ah, no. Sorry I was wrong...

shuheiktgw: Ahh wait, but what if `p.wasmRuntime.InstantiateModule` enqueues nil to `instanceModErrCh` and `writeCh` receives an error from `i.stdin.Write(cmd)`? In that case, don't we end up missing the error from `i.stdin.Write(cmd)`...?

goccy: `instanceModErrCh` returns an error when the module fails to start correctly (e.g., if `context.Context` has already been canceled ). `InstantiateModule` will continue to run until the plugin is no longer in use. So, we must call `(*CELPluginInstance).Close()` to send an exit signal to finish the wasm module. For this reason, `nil` is usually not returned from `instanceModErrCh`

shuheiktgw (approved):
LGTM 🙆‍♂️

PR #169, opened 2024-05-10
Fix cel plugin function for enum type

goccy:
Enable specifying an enum type for plugin functions as follows

```
enum EnumType {
  ENUM_FOO = 0;
  ENUM_BAR = 1;
}

option (grpc.federation.plugin).export = {
  name : "plugin"
  functions {name : "example" args {name : "v" type : "EnumType"} return {type : "bool"}}
};

```

### DIFF
--- a/resolver/cel.go
+++ b/resolver/cel.go
@@ -175,14 +175,14 @@ func (f *CELFunction) signatures(args []*Type, ret *Type) []*CELPluginFunctionSi
 		if arg.Kind == types.Enum {
 			sigs = append(sigs,
 				f.signatures(
-					append(append(append([]*Type{}, args[:idx]...), &Type{Kind: types.Int32}), args[idx+1:]...),
+					append(append(append([]*Type{}, args[:idx]...), Int32Type), args[idx+1:]...),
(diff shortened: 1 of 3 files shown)


resolver/cel.go
```diff
@@ -146,16 +147,64 @@ func (plugin *CELPlugin) LibraryName() string {
...
 
-func (f *CELFunction) CELReturn() *cel.Type {
-	return ToCELType(f.Return)
+func (f *CELFunction) signatures(args []*Type, ret *Type) []*CELPluginFunctionSignature {
+	var sigs []*CELPluginFunctionSignature
+	for idx, arg := range args {
+		if arg.Kind == types.Enum {
+			sigs = append(sigs,
+				f.signatures(
+					append(append(append([]*Type{}, args[:idx]...), Int32Type), args[idx+1:]...),
```
shuheiktgw, line 178: super-nit, `append(append(args[:idx], Int32Type), args[idx+1:]...)` could be more readable

shuheiktgw (approved):
LGTM 🙆‍♂️

Issue #160, opened 2024-04-22
Output gRPC Request contents by debug log


goccy:
For easier debugging

PR #170, opened 2024-05-10
Add debug log to output grpc request contents

utahta:
SSIA
https://github.com/mercari/grpc-federation/issues/160

### DIFF
--- a/generator/code_generator.go
+++ b/generator/code_generator.go
@@ -935,7 +935,14 @@ type LogValueAttr struct {
 }
 
 func (s *Service) LogValues() []*LogValue {
-	for _, msg := range s.Service.Messages {
+	msgs := s.Service.Messages
+	for _, dep := range s.Service.ServiceDependencies() {
+		for _, method := range dep.Service.Methods {
+			msgs = append(msgs, method.Request)
+		}
+	}
+
+	for _, msg := range msgs {
 		if msg.Rule != nil {
 			s.setLogValueByMessageArgument(msg)
 		}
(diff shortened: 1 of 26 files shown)


generator/templates/eval.go.tmpl
```diff
@@ -111,6 +111,9 @@ grpcfed.EvalDef({{ .Ctx }}, value, grpcfed.Def[{{ $def.Type }}, *localValueType]
 		}
 		{{- end }}
 		{{- end }}
+		{{- if $def.IsCall }}
+		s.logger.DebugContext(ctx, "call {{ $def.MethodFQDN }}", slog.Any("call_request", s.logvalue_{{ $def.LogValueRequestType }}(args)))
```
goccy, line 115: Here, I want to specify the FQDN of the request message

e.g.) `slog.Any("pkg.FooRequest", s.logvalue_Pkg_FooRequest(args))`

generator/code_generator.go
```diff
@@ -2454,6 +2461,17 @@ func (d *VariableDefinition) RequestType() string {
 	return ""
 }
 
+func (d *VariableDefinition) LogValueRequestType() string {
+	expr := d.VariableDefinition.Expr
+	switch {
+	case expr.Call != nil:
+		return fullMessageName(expr.Call.Request.Type)
+	case expr.Message != nil:
```
goccy, line 2469: Since it will always be called under `$def.IsCall` in template, is this condition actually not evaluated ?

utahta: Yes, this condition is not actually evaluated. I did the same as RequestType(), but we can remove it if not needed.

goccy: Sure, Let's remove it then !

goccy (approved):
LGTM !!!

PR #171, opened 2024-05-13
Fix oneof field of gRPC method request

goccy:
If specifying a oneof field during method invocation, it must always be specified by a conditional statement.
So, I added `if` feature for `def.call.request` .

resolve https://github.com/mercari/grpc-federation/pull/171


shuheiktgw (approved):
LGTM 🙆‍♂️

shuheiktgw (approved):
LGTM 🙆‍♂️

PR #174, opened 2024-05-13
Support SetLogger function

goccy:
Support `grpcfed.SetLogger` API for use with custom resolver

### DIFF
--- a/grpc/federation/context.go
+++ b/grpc/federation/context.go
@@ -6,8 +6,9 @@ import (
 )
 
 type (
-	loggerKey      struct{}
-	celCacheMapKey struct{}
+	loggerKey              struct{}
+	customResolverValueKey struct{}
+	celCacheMapKey         struct{}
 )
 
 func WithLogger(ctx context.Context, logger *slog.Logger) context.Context {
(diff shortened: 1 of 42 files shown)


grpc/federation/context.go
```diff
@@ -22,6 +23,35 @@ func Logger(ctx context.Context) *slog.Logger {
...
+	}
+	value.(*CustomResolverValue).Logger = logger
+}
+
+// CustomResolverValue type for storing references to be modified when calling a custom resolver.
+type CustomResolverValue struct {
+	Logger *slog.Logger
+}
+
+func WithCustomResolverValue(ctx context.Context) context.Context {
```
shuheiktgw, line 41: May I ask why we need this struct? If all we want to pass to custom resolvers is the logger, users can extract the logger directly from the context using `grpcfed.Logger`. Are we going to pass other values as well in the future?

goccy: That's right! I've made it a struct so it can be easily extended in the future.
Also, user cannot directly rewrite the logger instance within the custom resolver.

Issue #173, opened 2024-05-13
Add `sort` macro


goccy:
Add `range.sort(cur, next, <condition>)` to `grpc.federation.list` library

The example is follows.

```proto
def { name: "v" by: "[3, 1, 2]" }
def { by: "v.sort(cur, next, cur < next)" } // [1, 2, 3]
```

PR #176, opened 2024-05-17
Add sort macro

utahta:
- Add sort macro for list
  - `sortAsc(var, <expr>)`
  - `sortDesc(var, <expr>)`
  - `sortStableAsc(var, <expr>)`
  - `sortStableDesc(var, <expr>)`

resolve https://github.com/mercari/grpc-federation/issues/173


grpc/federation/cel/list.go
```diff
@@ -1,15 +1,26 @@
...
 	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/common/types/traits"
 	"github.com/google/cel-go/parser"
 )
 
 const ListPackageName = "list"
 
+const (
+	listSortAscFunc        = "grpc.federaiton.list.@sortAsc"
```
utahta, line 18: 📝  I intend the function with `@` prefix as a private. I referred it from [ext/math.go](https://github.com/google/cel-go/blob/master/ext/math.go#L338), but this is just a mark and does not appear to be a restriction.

grpc/federation/cel/list.go
```diff
@@ -61,6 +110,93 @@ func makeFirst(mef cel.MacroExprFactory, target ast.Expr, args []ast.Expr) (ast.
...
+	})
+	if hasErr {
+		return types.NewErr("cannot sort because some elements of list are not comparable")
+	}
+
+	resVals := make([]ref.Val, 0, len(vals))
+	for _, v := range vals {
+		resVals = append(resVals, v.OrigVal)
+	}
+	return types.DefaultTypeAdapter.NativeToValue(resVals)
```
goccy, line 197: I think it might be better to use `CELTypeHelper`, so how about passing `celTypeHelper` instance to `cel.NewLibrary()` and using it here?

https://github.com/mercari/grpc-federation/blob/main/grpc/federation/cel.go#L281-L282

grpc/federation/cel/list.go
```diff
@@ -61,6 +110,93 @@ func makeFirst(mef cel.MacroExprFactory, target ast.Expr, args []ast.Expr) (ast.
...
+	// The element being sorted must be values before expansion.
+	type sortVal struct {
+		OrigVal     ref.Val
+		ExpandedVal ref.Val
+	}
+	vals := make([]*sortVal, 0, int64(origLister.Size().(types.Int)))
+	for i := types.IntZero; i < origLister.Size().(types.Int); i++ {
+		origVal := origLister.Get(i)
+		expVal := expandedLister.Get(i)
+		if _, ok := expVal.(traits.Comparer); !ok {
```
goccy, line 166: 👍

grpc/federation/cel/list_test.go
```diff
@@ -64,11 +68,271 @@ func TestList(t *testing.T) {
...
+			expr: `[
+grpc.federation.cel.test.Message{id:"A", num:25}, 
+grpc.federation.cel.test.Message{id:"A", num:75}, 
+grpc.federation.cel.test.Message{id:"A", num:75}, 
+grpc.federation.cel.test.Message{id:"B", num:75},
+grpc.federation.cel.test.Message{id:"B", num:25},
+grpc.federation.cel.test.Message{id:"C", num:25},
+grpc.federation.cel.test.Message{id:"E", num:75},
+grpc.federation.cel.test.Message{id:"E", num:25},
+].sortStableDesc(v, v.num)`,
```
goccy, line 276: If you're testing `sortStableDesc`, I think it's better to run it twice and verify that the ids are also in descending order.

`[].sortStableDesc(v, v.num).sortStableDesc(v, v.id)`

goccy:
Please try using the sort feature around here and verify if it works with the existing message.
https://github.com/mercari/grpc-federation/blob/main/_examples/02_simple/proto/federation/federation.proto#L21-L52

grpc/federation/cel/list.go
utahta: Makes sense. I'll fix it!

grpc/federation/cel/list.go
goccy: I initially thought this method would be fine, but I realized that with this implementation, it cannot result in an error during the compile phase. Please add tests in `validator/validator_test.go` to check if it can produce an error

goccy (approved):
LGTM !!!!

PR #177, opened 2024-05-21
Fix auto type conversion

goccy:
- Enable automatic conversion of numeric types with validation
- Use alias for automatic conversion of request message
- Add an error to the return value of the type conversion function
- remove `validateMessageAlias` function as it is unnecessary


grpc/federation/cast.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func Int32ToUint32(v int32) (uint32, error) {
+	if v < 0 {
+		return 0, fmt.Errorf("failed to convert int32(%d) to uint32: %w", v, ErrOverflowTypeConversion)
+	}
+	return uint32(v), nil
+}
+
+func Int32ToUint64(v int64) (uint64, error) {
```
shuheiktgw, line 33: `v` should have been int32?

grpc/federation/cast.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func Uint64ToUint32(v uint64) (uint32, error) {
+	if (1 << 32) <= v {
+		return 0, fmt.Errorf("failed to convert uint64(%d) to uint32: %w", v, ErrOverflowTypeConversion)
+	}
+	return uint32(v), nil
+}
+
+func Uint32ToInt32(v uint64) (int32, error) {
```
shuheiktgw, line 61: `v` should have been uint32?

resolver/resolver.go
```diff
@@ -1002,6 +967,109 @@ func (r *Resolver) validateMessageFields(ctx *context, msg *Message, builder *so
 	}
 }
 
+func (r *Resolver) validateRequestFieldType(ctx *context, fromType *Type, toField *Field, builder *source.RequestOptionBuilder) {
+	if fromType == nil || toField == nil {
+		return
+	}
+	toType := toField.Type
+	if toType.Kind == types.Message {
```
shuheiktgw, line 975: Could toType be nil here?

shuheiktgw (approved):
LGTM 🙆‍♂️

resolver/resolver.go
goccy: It's OK, `toField` is one of the target gRPC method request message's field, so it is not nil always.

PR #181, opened 2024-05-23
Change schema of message and enum type to support multiple aliases

goccy:
Support multiple aliases for `grpc.federation.message` and `grpc.federation.enum` option.
When multiple aliases are specified, the resolution will use the elements common to each of them.


generator/code_generator.go
```diff
@@ -1932,7 +1943,22 @@ func (m *Message) autoBindFieldToReturnField(field *resolver.Field, autoBindFiel
...
-		fromType := field.SourceType()
+		var fromType *resolver.Type
+		for _, sourceType := range field.SourceTypes() {
+			if typ := m.autoBindSourceType(autoBindField, sourceType); typ != nil {
+				fromType = typ
+				break
+			}
+		}
+		if fromType == nil {
+			panic(
```
shuheiktgw, line 1954: Was it hard to return an error instead of panicking...?

shuheiktgw (approved):
LGTM 🙆‍♂️

generator/code_generator.go
goccy: Yes, this function cannot return an error because it is called when evaluating the template.

PR #182, opened 2024-05-24
Fix to resolve map

utahta:
SSIA


goccy:
Thank you for your awesome PR !
Could you add the case where the cast process of a map occurs an error to the validator package ?

resolver/resolver.go
```diff
@@ -3320,18 +3354,54 @@ const privateProtoFile = "grpc/federation/private.proto"
...
 	for idx, field := range arg.Fields {
-		var typeName string
-		if field.Type.Message != nil {
+		var (
+			typeName string
+			label    = descriptorpb.FieldDescriptorProto_LABEL_OPTIONAL
+		)
+		if field.Type.Message != nil && field.Type.Message.IsMapEntry {
+			mp := messageArgumentDescriptor(field.Type.Message)
+			if mp.GetName() == "" {
```
goccy, line 3393: I didn't expect it to be empty, but under what circumstances would it be empty ?

goccy: I understand ! OK 👍

goccy (approved):
LGTM 👍

PR #183, opened 2024-05-24
Fix message argument type conversion

goccy:
For numeric types, the specification allows accepting them even if the type of the message argument differs. In such cases, since the actual message argument type may differ, additional type conversion is necessary.


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #186, opened 2024-05-30
Support overload for cel plugin functions

goccy:
- Add suffix to function name to support overload feature for CEL plugin
- Delete wasm raw file and generates it at testing


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #187, opened 2024-06-03
Fix the code generation for names that include numbers

goccy:
SSIA


shuheiktgw:
Should we update [here](https://github.com/mercari/grpc-federation/blob/be99e50be7b1d923ff2037c08a60d11b395731fa/generator/code_generator.go#L1825C36-L1825C54) as well?

shuheiktgw (approved):
LGTM 🙆‍♂️

PR #189, opened 2024-06-04
Support wrapper types in google.protobuf package

goccy:
cel-go treats the types defined in `google.protobuf.wrappers` in a special way, assigning them a kind of primitive type. However, in gRPC Federation, it is necessary to identify wrapper types because we want to handle the original message type during type conversion.


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #191, opened 2024-06-05
Support null value

goccy:
Support `null` literal in CEL.

- Add `IsNull` field to `resolver.Type` to distinct null typed value


utahta (approved):

shuheiktgw:
I'm worried about the following two lines. If `types.ToKind` is called against the new null type, it should return `types.Unknown` and the code flow may not be right...
- https://github.com/mercari/grpc-federation/blob/836f108277aba84d3c68b0d8740620f352ce1f9b/resolver/resolver.go#L465-L476
- https://github.com/mercari/grpc-federation/blob/3024c00326f77fdeef88954c80bf41fc68e4dc67/lsp/server/definition.go#L85-L100

goccy:
@shuheiktgw Since null type is only created as a result of evaluating CEL, it cannot be used in the definition of plugin functions or field types. Therefore, I think that null type will be not evaluated in these contexts.

PR #194, opened 2024-06-10
Remove consts

shuheiktgw:
SSIA. Since we decided to remove the constant fields and use `by` instead, I removed those fields and related lines.


validator/validator_test.go
```diff
@@ -680,7 +684,7 @@ testdata/invalid_message_map_alias.proto:37:3: cannot convert type automatically
 				}
 				return
 			}
-
+			fmt.Println(got)
```
goccy, line 687: Is this debugging code ?

shuheiktgw: 🙇

goccy (approved):
Thank you for your awesome PR !!! LGTM 👍

Issue #172, opened 2024-05-13
Support `grpc.federation.log` library


goccy:
Add a CEL library for manipulating `slog.Logger`. On the proto side, We'd like to add fields for logger and be able to log at any desired logging level anywhere (mainly for adding `Info` logs that we want to display in production environment )

PR #195, opened 2024-06-11
Add grpc.federation.log library

utahta:
Add grpc.federation.log library for manipulating slog.Logger.
https://github.com/mercari/grpc-federation/issues/172


goccy (approved):
LGTM 👍

PR #198, opened 2024-06-19
Fix presence detection of field's type value

utahta:
In message argument types, there are cases where has(e.f) should be false but is true.
I think we need to check if the value is a zero value when detecting the presence of field type value.


goccy (approved):
Thank you for your awesome PR !! 🙏

PR #200, opened 2024-06-24
Add enum selector

goccy:
- Treats enum identifier as `opaque<int>` type at compile time
- Refactor generator templates to add the enum selector process
- Add `grpc.federation.enum.select` function to select different enum values for using field binding by multiple aliases


shuheiktgw (approved):
LGTM 🙆‍♂️

utahta (approved):

PR #202, opened 2024-06-25
Fix enum selector

goccy:
As a result of processing all enum values as `opaque<int>`, it was found that type errors occur when trying to create a message within a CEL expression using those values.

This behavior was unintended, so we will revert to processing them as int as before and convert them to `opaque<int>` only when necessary.


resolver/resolver.go
```diff
@@ -3357,20 +3390,6 @@ func (r *Resolver) resolveCELValue(ctx *context, env *cel.Env, value *CELValue)
 	if issues.Err() != nil {
 		return errors.Join(append(r.celRegistry.errors(), issues.Err())...)
 	}
-	if len(r.celRegistry.usedEnumValueMap) != 0 {
```
shuheiktgw, line 3360: It seems we don't need the `usedEnumValueMap` field to `CELRegistry`?

shuheiktgw (approved):

resolver/resolver.go
```diff
@@ -1292,6 +1304,27 @@ func (r *Resolver) validateBindFieldType(ctx *context, fromType *Type, toField *
...
+		return nil
+	}
+
+	if fromType == nil {
+		return nil
+	}
+	if fromType.Kind == types.Message && fromType.Message != nil && fromType.Message.IsEnumSelector() {
+		return nil
+	}
+	return errors.New(`if multiple aliases are specified, you must use grpc.federation.enum.select function to bind`)
```
utahta, line 1325: Is it possible to add a test case for this error?

resolver/resolver.go
goccy: That's correct ! However, since we would like to eventually implement a proper mechanism for handling enum value, I plan to keep the mechanism in place.

utahta (approved):
LGTM!

PR #208, opened 2024-07-05
Fix error log

utahta:
Output log attributes defined in proto (using `grpc.federation.log.add`) when validation and gRPC call errors.


grpc/federation/error.go
```diff
@@ -38,18 +40,53 @@ func GoWithRecover(eg *errgroup.Group, fn func() (any, error)) {
...
+
+func (e *ErrorWithLogAttrs) Error() string {
+	return e.err.Error()
+}
+
+func (e *ErrorWithLogAttrs) Unwrap() error {
+	return e.err
+}
+
+func (e *ErrorWithLogAttrs) GRPCStatus() *grpcstatus.Status {
```
shuheiktgw, line 56: May I ask where this is going to be used? 🤔

grpc/federation/error.go
```diff
@@ -38,18 +40,53 @@ func GoWithRecover(eg *errgroup.Group, fn func() (any, error)) {
...
+	return grpcstatus.Convert(e.err)
+}
+
+func NewErrorWithLogAttrs(err error, attrs []slog.Attr) error {
+	if err == nil {
+		return nil
+	}
+	if errors.Is(err, &ErrorWithLogAttrs{}) {
```
shuheiktgw, line 64: If I use this function, when the error is `ErrorWithLogAttrs,` I might want the function to merge `attrs` instead of simply returning the original error.

goccy (approved):
LGTM 👍

grpc/federation/error.go
utahta: This is used from [grpcstatus.FromError](https://github.com/mercari/grpc-federation/blob/27cf04fb39a1971fd495481a2d48e7add23b22b4/grpc/federation/error.go#L84).
By the way, I originally thought that I only needed to implement Unwrap method, but in fact the error message changed unintentionally, so I implemented this method as well.

grpc/federation/error.go
utahta: I considered the case where an error is wrapped multiple times in a nested message, but I removed this code because such a case does not currently exist and it would be better to check the error outside of this function. 🙏

goccy (approved):

shuheiktgw (approved):
👍

PR #210, opened 2024-07-18
Support multi env definitions in one file

shuheiktgw:
Allow users to define multiple environment variables in the same package. The result env message will be the intersection set of each service's environment variables.


resolver/resolver.go
```diff
@@ -3507,33 +3494,80 @@ func (r *Resolver) createCELEnv(msg *Message) (*cel.Env, error) {
...
+		}
+		if omit {
+			continue
+		}
+
+		// Check if all the envVar has the same type
+		typ := svcEnvVars[0].envVar.Type
+		var mismatch bool
+		for _, sev := range svcEnvVars[1:] {
+			if typ.Kind != sev.envVar.Type.Kind {
```
shuheiktgw, line 3549: This comparison could be wrong if the type is repeated or map, but I'm not too sure what is the best way to compare the two types. Should I compare each fields in Type instead? 🤔

validator/validator_test.go
```diff
@@ -90,12 +90,10 @@ testdata/invalid_env.proto:24:16: "org.federation.Invalid" message does not exis
...
 `},
 		{file: "invalid_multiple_env.proto", expected: `
-testdata/invalid_multiple_env.proto:12:9: "org.federation.InlineEnvService": multiple services within the same package ("org.federation") cannot use the env option
-12:      env {
-             ^
-testdata/invalid_multiple_env.proto:48:9: "org.federation.RefEnvService": multiple services within the same package ("org.federation") cannot use the env option
-48:      env { message: "Env" }
-             ^
+testdata/invalid_multiple_env.proto:56:1: environment variable "ccc" has different types across services: service: InlineEnvService, type: int64, repeated: false, service: RefEnvService, type: int64, repeated: true
+environment variable "eee" has different types across services: service: InlineEnvService, type: message, repeated: false, service: RefEnvService, type: message, repeated: false
```
shuheiktgw, line 94: The current error message does not show which field does not match, but I could not think of an easy way to do that...

goccy: I think it's sufficient to just display the name of the Env and the different Service names.

resolver/resolver.go
```diff
@@ -3507,33 +3494,85 @@ func (r *Resolver) createCELEnv(msg *Message) (*cel.Env, error) {
...
+			}
+			// Fix the output order mainly for testing
+			sort.Strings(attrs)
+			errs = append(errs, fmt.Errorf("environment variable %q has different types across services: %s", envVarName, strings.Join(attrs, ", ")))
+		}
+
+		envVars = append(envVars, svcEnvVars[0].envVar)
+	}
+	if len(errs) > 0 {
+		return nil, errors.Join(errs...)
```
goccy, line 3572: It might become difficult to read if there are multiple errors, so it seems better to add each error to the context using addError.

resolver/resolver.go
```diff
@@ -3507,33 +3494,85 @@ func (r *Resolver) createCELEnv(msg *Message) (*cel.Env, error) {
...
-			}
-			if svc.Rule.Env == nil {
-				continue
+
+	// Examine all the services where the message is used and collect all the essential information
+	// to calculate the intersection set of each service's environment variables.
+	svcEnvVarMap := map[*Service]map[string]*EnvVar{}
+	envVarNameSet := map[string]struct{}{}
+	for svc, msgSet := range svcMsgSet {
+		if _, exist := msgSet[msg]; exist {
```
goccy, line 3510: nits: Since `exists` is used in all other places, please align this one as well if possible.

goccy (approved):
LGTM 👍

shuheiktgw:
🚀

PR #212, opened 2024-07-26
Add import rule

utahta:
Add `grpc.federation.file.import` rule to avoid `unused import statement` lint errors when importing for the purpose of resolving references to methods used by the federation.


proto/grpc/federation/federation.proto
```diff
@@ -43,6 +43,7 @@ extend google.protobuf.OneofOptions {
 
 message FileRule {
   CELPlugin plugin = 1;
+  repeated string import = 2;
```
goccy, line 46: Please add the documentation for the import feature here 🙏

goccy (approved):

PR #225, opened 2024-08-13
Pass context.Context to CEL function

goccy:
- Use CEL Macros to pass context.Context argument to CEL function
    - e.g.) `pkg.func(arg)` => `pkg.func(grpc.federation.private.Context, arg)`
- Use singleton instance for all CEL libraries
    - Remove `UseContextLibrary` flag
    - wasm plugins are initialized at constructor


shuheiktgw:
IIUC, [this field](https://github.com/mercari/grpc-federation/blob/1ee7cffcd790c8fc4bdd637c98ef61183477bc8e/grpc/federation/lib.go#L86) should be removed as well?

shuheiktgw (approved):

PR #227, opened 2024-08-15
Improve scheduler

goccy:
Make `validation` priority to the highest.
If we write the following options, the order is `validationA` => ( `callA` , `validationB` ) => `callB` .
(※ `callA` and `validationB` are called concurrently )

```proto
def { validation {} } // validationA
def { call {} } // callA
def { validation {} } // validationB
def { call {} } // callB
```

The important thing is that `call` block is executed after `validation` block.


utahta (approved):
LGTM!

PR #228, opened 2024-08-16
Add configuration to output logs as any log level

utahta:
Add configuration to output logs as any log level in the error block.


goccy (approved):
LGTM 👍

PR #234, opened 2024-08-20
Fix Candidates method file match

shuheiktgw:
`strings.HasSuffix(file.GetName(), loc.FileName)` could match the wrong file if there are multiple files with the same name. Instead, set the fileName field to the full file path and use that.


goccy (approved):
LGTM 👍

shuheiktgw:
🙏

PR #235, opened 2024-08-21
Fix time api

goccy:
Since `google.protobuf.Timestamp` does not have time zone information, when we use `withLocation()` API, the time zone information may be lost if it bypass Go's value ( `timestamppb.Timestamp` ).
So, I fix this issue with custom `Time` message ( it keeps time zone information ).


shuheiktgw (approved):
LGTM 🙆‍♂️

PR #236, opened 2024-08-21
Fix conflicted alias name

goccy:
If there are same package name, build error occurs, so I add the suffix to package name to make unique name.


shuheiktgw (approved):
LGTM :ok

utahta (approved):

PR #240, opened 2024-08-23
Show warning about file import

utahta:
Show the warning in the following cases:
- If proto file is imported by grpc.federation.file.import statement and unused for the definition of grpc federation:
     - `Import example.proto is unused for the definition of grpc federation.`
- If proto file is imported by protobuf statement and used only for the definition of grpc federation:
     - `Import example.proto is used only for the definition of grpc federation. You can use grpc.federation.file.import instead.`


goccy (approved):
Awesome !!! LGTM 👍

PR #242, opened 2024-09-03
Support custom response message

goccy:
Spec: https://github.com/mercari/grpc-federation/blob/support-custom-response/docs/references.md#grpcfederationmethodresponse


utahta (approved):
LGTM!

PR #243, opened 2024-09-06
Cast null_value

utahta:
Evaluates an expression such as `msg == null` by converting it to `grpc.federation.cast.null_value(msg) == null`.
For example, when there is an expression `msg == null`, if nil is directly assigned to msg, there is no problem, but if typed nil, there is an unintended behavior (in this case, false is returned).


grpc/federation/cel.go
```diff
@@ -832,3 +859,169 @@ func SetCELValue[T any](ctx context.Context, param *SetCELValueParam[T]) error {
...
+
+func (r *nullValueFuncReplacer) Replace(checkedExpr *exprpb.CheckedExpr) bool {
+	r.init(checkedExpr)
+	newExprVisitor().Visit(checkedExpr.GetExpr(), func(e *exprpb.Expr) {
+		switch e.GetExprKind().(type) {
+		case *exprpb.Expr_CallExpr:
+			r.replaceCall(e)
+		case *exprpb.Expr_ComprehensionExpr:
+			// Comprehension is not supported by parser.Unparse.
+			r.unsupported = true
```
goccy, line 907: If a comparison with `null` is made while using comprehension, the comparison with `null` will be performed without being replaced. Is it okay not to treat this as an error ?

utahta: I see. I will try to fix it so that it parses once on the resolver side and returns a validation error if comprehension is used.

utahta: This exception handling is not needed and I removed it.
Thank you for pointing it out 🙏

goccy (approved):
Awesome!! LGTM 👍

PR #244, opened 2024-09-17
Support URL CEL library

pikachu0310:
Provide functionalities similar to Go's [net/url](https://pkg.go.dev/net/url) using the grpc.federation.url library.

- [x] impl url.parse
- [x] impl URL functions
- [x] impl Userinfo functions
- [x] impl other functions
- [x] add url.parse test
- [x] add URL functions test
- [x] add Userinfo functions test
- [x] add other functions test
- [x] write docs/cel/url.md


grpc/federation/cel/url.go
```diff
@@ -0,0 +1,611 @@
...
+	return url.URL{
+		Scheme:     x.GetScheme(),
+		Opaque:     x.GetOpaque(),
+		User:       user,
+		Host:       x.GetHost(),
+		Path:       x.GetPath(),
+		RawPath:    x.GetRawPath(),
+		ForceQuery: x.GetForceQuery(),
+		RawQuery:   x.GetRawQuery(),
+		Fragment:   x.GetFragment(),
```
shuheiktgw, line 39: Would you explain why you omit some fields, such as OmitHost and RawFragment, here?

grpc/federation/cel/url.go
```diff
@@ -0,0 +1,611 @@
...
+	}, nil
+}
+
+func (x *Userinfo) GoUserinfo() (*url.Userinfo, error) {
```
shuheiktgw, line 43: According to the implementation, GoUserinfo won't return nil, but is it really correct? For example, should we return nil when both username and password are empty? The original UserInfo struct assumes the case when u (*UserInfo) is nil and returns an empty string.

grpc/federation/cel/url.go
```diff
@@ -0,0 +1,611 @@
...
+				func(_ context.Context, self ref.Val, args ...ref.Val) ref.Val {
+					v, err := self.Value().(*URL).GoURL()
+					if err != nil {
+						return types.NewErr(err.Error())
+					}
+					return types.String(v.RawPath)
+				},
+			),
+		),
+		// forceQuery is changed to return String
```
shuheiktgw, line 270: Why is it? 🤔 Would you add a comment to it?

grpc/federation/cel/url.go
```diff
@@ -0,0 +1,611 @@
...
+		BindMemberFunction(
+			"password",
+			MemberOverloadFunc(createURLID("password_userinfo_string"), UserinfoType, []*cel.Type{}, cel.StringType,
+				func(_ context.Context, self ref.Val, args ...ref.Val) ref.Val {
+					v, err := self.Value().(*Userinfo).GoUserinfo()
+					if err != nil {
+						return types.NewErr(err.Error())
+					}
+					password, hasPassword := v.Password()
+					if hasPassword {
```
shuheiktgw, line 569: nit, Technically, this is different from the original implementation. The original implementation returns `"", false` only when u (*UserInfo) is nil. It does not check whether hasPassword is true or not.

shuheiktgw (commented):
I left some comments, so PTAL 🙏

grpc/federation/cel/url.go
pikachu0310: I made a mistake and have now corrected it.

grpc/federation/cel/url.go
pikachu0310: This issue was caused by an implementation mistake that occurred in the past. It has been corrected now.

grpc/federation/cel/url.go
pikachu0310: I misread the implementation. I have corrected it.

grpc/federation/cel/url.go
pikachu0310: It has been fixed.

pikachu0310:
I have addressed ktgw's review comments and resolved the conflicts. Please review it again.

grpc/federation/cel/url.go
```diff
@@ -0,0 +1,621 @@
...
+	return lib.typeAdapter.NativeToValue(&URL{
+		Scheme:     v.Scheme,
+		Opaque:     v.Opaque,
+		User:       userinfo,
+		Host:       v.Host,
+		Path:       v.Path,
+		RawPath:    v.RawPath,
+		ForceQuery: v.ForceQuery,
+		RawQuery:   v.RawQuery,
+		Fragment:   v.Fragment,
```
shuheiktgw, line 103: The same as https://github.com/mercari/grpc-federation/pull/244#discussion_r1777962091

grpc/federation/cel/url.go
```diff
@@ -0,0 +1,621 @@
...
+		RawPath:     x.GetRawPath(),
+		OmitHost:    x.GetOmitHost(),
+		ForceQuery:  x.GetForceQuery(),
+		RawQuery:    x.GetRawQuery(),
+		Fragment:    x.GetFragment(),
+		RawFragment: x.GetRawFragment(),
+	}, nil
+}
+
+func (x *Userinfo) GoUserinfo() (*url.Userinfo, error) {
```
shuheiktgw, line 45: I believe we don't need to return an error since there is no case to return it? 🤔

grpc/federation/cel/url.go
```diff
@@ -0,0 +1,621 @@
...
+					if err != nil {
+						return types.NewErr(err.Error())
+					}
+					return types.Bool(v.OmitHost)
+				},
+			),
+		),
+		BindMemberFunction(
+			"forceQuery",
+			MemberOverloadFunc(createURLID("forceQuery_url_string"), URLType, []*cel.Type{}, cel.BoolType,
```
shuheiktgw, line 288: I believe it should be `forceQuery_url_bool`.

shuheiktgw (commented):
I left some comments so PTAL 🙏

shuheiktgw (approved):
LGTM

grpc/federation/cel/url.go
```diff
@@ -0,0 +1,528 @@
...
+					v := lib.refToGoURLValue(self)
+					return types.String(v.Scheme)
+				},
+			),
+		),
+		BindMemberFunction(
+			"opaque",
+			MemberOverloadFunc(createURLID("opaque_url_string"), URLType, []*cel.Type{}, cel.StringType,
+				func(_ context.Context, self ref.Val, args ...ref.Val) ref.Val {
+					v := self.Value().(*URL).GoURL()
```
utahta, line 207: Should we use `lib.refToGoURLValue` instead? (the same goes for other code as well)

pikachu0310: I forgot about that! I have corrected it.

PR #245, opened 2024-09-20
Support Strings CEL library

pikachu0310:
Provide functionalities similar to Go's [strings](https://pkg.go.dev/strings) using the grpc.federation.strings library.
https://github.com/mercari/grpc-federation/issues/159


grpc/federation/cel/strings.go
```diff
@@ -0,0 +1,405 @@
...
+		),
+		BindFunction(
+			createStringsName("count"),
+			OverloadFunc(createStringsID("count_string_string_int"), []*cel.Type{cel.StringType, cel.StringType}, cel.IntType,
+				func(_ context.Context, args ...ref.Val) ref.Val {
+					return types.Int(strings.Count(args[0].(types.String).Value().(string), args[1].(types.String).Value().(string)))
+				},
+			),
+		),
+		// func Cut(s, sep string) (before, after string, found bool)
```
shuheiktgw, line 95: What is this comment? 🤔

grpc/federation/cel/strings.go
```diff
@@ -0,0 +1,405 @@
...
+			),
+		),
+		BindFunction(
+			createStringsName("indexByte"),
+			OverloadFunc(createStringsID("indexByte_string_byte_int"), []*cel.Type{cel.StringType, cel.BytesType}, cel.IntType,
+				func(_ context.Context, args ...ref.Val) ref.Val {
+					return types.Int(strings.IndexByte(args[0].(types.String).Value().(string), args[1].(types.Bytes).Value().(byte)))
+				},
+			),
+			OverloadFunc(createStringsID("indexByte_string_string_int"), []*cel.Type{cel.StringType, cel.StringType}, cel.IntType,
```
shuheiktgw, line 181: Would you explain why we need this `OverloadFunc` even if we already have the index function? 🙏

shuheiktgw (commented):
I left some questions but overall it looks good to me 👍

grpc/federation/cel/strings.go
pikachu0310: I referred to the implementation linked below. While testing, I encountered an issue when trying to pass a Go `byte` type as an argument. It seems that trying to retrieve it using `cel.bytes` didn't work as expected. By defining the argument as `cel.string` instead, I was able to successfully pass and retrieve the value.

https://github.com/mercari/grpc-federation/blob/b948f38603ce28df91b0b67cd3f0af65b54f1ec4/grpc/federation/cel/time.go#L582-L612

grpc/federation/cel/strings.go
pikachu0310: Sorry for the lack of explanation in the comments. I left a comment for the functions I determined to be impossible to implement. Except for `Cut`, the other comments are for functions that have arguments of types (like Go's `func` type) that cannot be received by CEL. As for `Cut`, I concluded it was impossible to implement because it needs to return two variables. 

Is there any way to return two variables in this case?

shuheiktgw: We talked on this point directly and decided to update the comments to clarify why we cannot implement those functions.

grpc/federation/cel/strings.go
shuheiktgw: Makes sense!

shuheiktgw (approved):
LGTM 👍 
Waiting for an approval from @utahta.

grpc/federation/cel/strings.go
```diff
@@ -0,0 +1,405 @@
...
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+)
+
+const StringsPackageName = "strings"
+
+var _ cel.SingletonLibrary = new(StringsLibrary)
+
+type StringsLibrary struct {
+	typeAdapter types.Adapter
```
utahta, line 20: It seems that typeAdapter is not being used.  Is it possible to remove it?

pikachu0310: You're right, it's not being used. I've removed it. Thank you for pointing it out.

grpc/federation/cel/strings.go
pikachu0310: > Do we miss the Cut function?

Regarding "Cut," I found that it's feasible to implement, so I added the implementation.

utahta (approved):
LGTM!

shuheiktgw (approved):
🙆‍♂️

PR #250, opened 2024-12-02
Fix gRPC default error value

goccy:
### Breaking Changes

Currently, if neither code nor message is specified in `error {}`, default values are used. However, this could lead to unintended results, so if they are not specified, the original error values will be used instead.


kodams (approved):

generator/templates/error_handler.go.tmpl
```diff
@@ -4,6 +4,27 @@
...
+	defaultDetails = make([]grpcfed.ProtoMessage, 0, len(details))
+	for _, detail := range details {
+		msg, ok := detail.(grpcfed.ProtoMessage)
+		if ok {
+			defaultDetails = append(defaultDetails, msg)
+		}
+	}
+	_ = defaultMsg
+	_ = defaultCode
+	_ = defaultDetails
```
utahta, line 25: What is this for?

goccy: To prevent build errors for unused variables

utahta (approved):

PR #259, opened 2025-01-21
Support math CEL library

keitaj:
Support Golang's math package for CEL library
https://pkg.go.dev/math


goccy (approved):
Thank you for your contribution !!! LGTM 👍

PR #260, opened 2025-01-23
Fix generator for enum attribute

goccy:
- Fix generator for enum attribute
- Use early return for proto's compile error
- Refactor process of getting all variable definitions from message
- Fix unused import statement feature for enum attributes


validator/validator_test.go
```diff
@@ -758,12 +758,6 @@ invalid_message_map_alias.proto:39:3: cannot convert type automatically: map val
        ^
 `},
 		{file: "invalid_file_import.proto", expected: `
-invalid_file_import.proto:6:8: [WARN] Import post.proto is used only for the definition of grpc federation. You can use grpc.federation.file.import instead.
-6:  import "post.proto";
-           ^
-invalid_file_import.proto:12:5: [WARN] Import user.proto is unused for the definition of grpc federation.
-12:      "user.proto",
-         ^
```
utahta, line 766: It seems that tests for unused import warnings do not work because it now checks the error from `r.resolveFileImportRule`. [ref](https://github.com/mercari/grpc-federation/pull/260/files#diff-732f365bd38694dea48332c13814847ee19a76bebc5f3d2c87733f1a4ee915ffR160)
I think It would be good to split the test files into `invalid_file_unused_import.proto` and `invalid_file_unknown_import.proto`.

goccy: Thank you for reviewing ! I've fixed this point.

utahta (approved):

PR #278, opened 2025-03-04
Fix version text with `go install`

goccy:
- Use `debug.ReadBuildInfo()` for `go install`
- Fix default version text ( `dev` => `(devel)` )
- Add version command to `protoc-gen-grpc-federation`


utahta (approved):
👍

PR #292, opened 2025-05-21
Support wasi-socket for CEL plugin

goccy:
When defining a plugin, specifying the `capability` as follows allows the wasm plugin to access external resources.

```proto
option (grpc.federation.file).plugin.export = {
  name: "net"
  capability {
    network: {} // enable network access via http/https
    env { names: ["foo"] } // enable access to `FOO` environment variable
    file_system { mount_path: "/" } // enable access to file system from mounted point by `/`
  }
}
```


utahta (approved):
🎉

PR #293, opened 2025-05-22
Variables with the same name but different types cause a compilation error

utahta:
Split variable definitions into a separate struct and embed it to avoid compilation errors caused by name conflicts.


goccy (approved):
Thank you for your awesome PR !!! LGTM 👍

PR #319, opened 2025-08-22
Fix cel-ext libs

goccy:
- Support ex.tTwoVarComprehension library
- Support ext.Strings library
  - If there is already a function with the same name defined in the gRPC Federation standard library, such as `replace` or `format`, the macro expansion cannot determine whether it should pass context.Context as the first argument, and always passing it results in a compilation error. To address this, we always pass context.Context, and then use Eval further down the line to call the API of the external library.


utahta (approved):
LGTM!

PR #348, opened 2025-12-03
Add CLI ( grpcfedctl ) to create wasm cache directly

goccy:
Support the following command

```
# create wazero's compilation cache to `cache` directory
grpcfedctl cache create --name regexp --out=cache path/to/regexp.wasm 


# print version
grpcfedctl version
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,10 @@ generate/examples/%:
 	$(MAKE) -C $* generate
 
 .PHONY: build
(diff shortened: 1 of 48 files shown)


utahta (approved):

utahta:
📝 Just to summarize our offline discussion for the record:
- Regarding code reuse of cache generation: We want to keep the API internal at this stage (e.g., avoiding exposure under grpc/federation/cel), so we opted not to refactor for reuse yet.
- Regarding documentation: This command is currently intended for specific tooling needs (like internal service). Since it’s not a common workflow for general users, we’ll defer the documentation updates.

PR #351, opened 2026-01-08
Support parse UUID function

atkin-mc:
Add a uuid.parse function that calls the corresponding Golang uuid package function.


goccy (approved):
Thank you for your contribution !!

LGTM 👍

PR #356, opened 2026-01-20
Add switch to federation.proto

atkin-mc:
Add SwitchExpr to federation.proto. A `switch` is similar to the Golang switch statement. It can be used in service variable definitions as well as message variable definitions.

### DIFF
--- a/grpc/federation/federation.pb.go
+++ b/grpc/federation/federation.pb.go
@@ -1891,6 +1891,8 @@ type SwitchCaseExpr struct {
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
+	// if specify the condition for evaluating expr.
+	// this value evaluated by CEL and it must return a boolean value.
 	If string `protobuf:"bytes,1,opt,name=if,proto3" json:"if,omitempty"`
 	// expr specify the value to return for the case.
 	//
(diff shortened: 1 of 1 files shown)


proto/grpc/federation/federation.proto
```diff
@@ -351,6 +357,38 @@ message CallExpr {
...
+message SwitchExpr {
+  // Cases for the switch expression.
+  repeated SwitchCaseExpr case = 1;
+  // The default case, if none of the "case.if" expressions evaluate to true.
+  SwitchDefaultExpr default = 2;
+}
+
+// SwitchCaseExpr represents a single case for a switch expression.
+message SwitchCaseExpr {
+  string if = 1;
```
goccy: Please add the comment for this field

proto/grpc/federation/federation.proto
```diff
@@ -351,6 +357,38 @@ message CallExpr {
...
+  oneof expr {
+    string by = 11;
+  }
+}
+
+// SwitchDefaultExpr represents the default case for a switch expression.
+message SwitchDefaultExpr {
+  // expr specify the value to return for the default case.
+  oneof expr {
+    string by = 11;
```
goccy: Please add the comment for this field

atkin-mc: Added.

goccy (approved):

PR #360, opened 2026-02-10
Add variable definition "def" in switch

atkin-mc:
Add local variable definition to the `message.switch` statement. This follows the same rules of scoping as `message.validation.error` (i.e., all variables are in the same namespace regardless of scope, so it is not possible to introduce a duplicate variable name in a child scope). Variable definitions for a switch case are only evaluated when the `if` for the case evaluates to `true`.

This PR also adds test cases for `switch` to `source/file_test.go` and adds some functionality to `source/file.go` that was left out of https://github.com/mercari/grpc-federation/pull/356

Also, `docs/reference.md` doesn't mention `message.validation.error.def`, `message.validation.error.details.def`, this PR documents those fields.


goccy (approved):
Awesome !!! LGTM 👍

PR #361, opened 2026-02-25
Support proto3 optional fields

atkin-mc:
Support proto3 optional fields in messages. This requires (1) recognizing and recording that the field is optional; (2) correctly setting it in the generated code -- optional fields are translated into pointers in generated `pb.go`, so the code generator must distinguish them.

Recognizing an optional field is complicated by how optional fields are converted handled by the protoc-gen-go compiler: an optional field is treated (at the descriptor level) as oneof with a single field, and a hidden oneof descriptor. This "secret oneof" confuses the resolver, so special handling is needed to make sure that an optional field isn't interpreted as a oneof field.

Additionally, this PR makes the resolver reject an optional non-message field with a custom field resolver. The reason is that supporting this requires changing the resolver interface to return a pointer. This is a reasonable extension, but I am leaving it out of this PR to prevent it growing larger.


resolver/resolver.go
```diff
@@ -1667,6 +1673,18 @@ func (r *Resolver) validateMessageFields(ctx *context, msg *Message, builder *so
 			continue
 		}
 		if field.HasMessageCustomResolver() || field.HasCustomResolver() {
+			// TODO: custom_resolver on proto3 optional scalar/enum fields is not yet supported.
+			// To support it, ReturnType() in code_generator.go needs to return a pointer type
+			// (e.g. *int64 instead of int64) when the field is Proto3Optional, and the
+			// setReturnValueByFieldCustomResolver template needs to handle the pointer assignment.
+			if field.HasCustomResolver() && field.Proto3Optional && field.Type.Kind != types.Message {
```
utahta: I noticed that `field.Type` is accessed here, while the null check for it is performed just below this. Could this potentially cause an issue if `field.Type` is actually null?

generator/code_generator.go
```diff
@@ -2518,11 +2520,12 @@ func (m *Message) autoBindFieldToReturnField(field *resolver.Field, autoBindFiel
...
-		Value:        value,
-		RequiredCast: requiredCast,
-		CastFunc:     castFunc,
-		ProtoComment: fmt.Sprintf(`// { name: %q, autobind: true }`, name),
+		Name:           fieldName,
+		Value:          value,
+		RequiredCast:   requiredCast,
+		CastFunc:       castFunc,
+		ProtoComment:   fmt.Sprintf(`// { name: %q, autobind: true }`, name),
+		Proto3Optional: field.Proto3Optional && field.Type.Kind != types.Message,
```
utahta: This condition breaks for `optional bytes`.

protoc-gen-go keeps `optional bytes` as `[]byte` (not `*[]byte`) because `nil` already serves as the "unset" marker — unlike other scalars. With the current check, the template emits `ret.B = &v` against a `[]byte` field, which won't compile.

Quick repro:
```proto
optional bytes  b = 1;  // → B []byte
optional string s = 2;  // → S *string
```

`types.Bytes` should be excluded alongside `types.Message`:

```go
switch f.Type.Kind {
case types.Message, types.Bytes:
    return false
}
```

Could we also cover `optional bytes` in the example / expected test file?
Otherwise this path stays untested.

utahta: Since `field.Proto3Optional && field.Type.Kind != types.Message` appears in two places and likely needs more exceptions (bytes at minimum), could we extract it to a method on `resolver.Field`?

```go
// RequiresOptionalPointerWrap reports whether grpc-federation needs to
// wrap the generated value in a pointer for proto3 `optional`. Message
// fields are excluded because they are already pointers in generated Go;
// bytes fields are excluded because protoc-gen-go keeps them as []byte.
func (f *Field) RequiresOptionalPointerWrap() bool {
    if !f.Proto3Optional || f.Type == nil {
        return false
    }
    switch f.Type.Kind {
    case types.Message, types.Bytes:
        return false
    }
    return true
}
```

Keeping `Field.Proto3Optional` as the proto-level fact and introducing `RequiresOptionalPointerWrap()` as the code-gen judgment makes the two concepts distinct — which matters because `Proto3Optional` is `true` for optional messages too, while the wrap decision should be `false` there.

generator/templates/server.go.tmpl
```diff
@@ -872,13 +881,19 @@ ret.{{ .Name }} = {{ .Value }} {{ .ProtoComment }}
...
 if err != nil {
 	return err
 }
+{{- if .Proto3Optional }}
+ret.{{ $name }} = &{{ toLocalVariable $name }}
+{{- else }}
 ret.{{ $name }} = {{ toLocalVariable $name }}
+{{- end }}
+{{- else if .Proto3Optional }}
+ret.{{ $name }} = &{{ $v }}
```
utahta: I may be missing something, but I'm not sure what happens when an optional field on the request is propagated to an optional field on the response. A small example:

```proto
message GetPostRequest  { optional int64 score = 1; }
message GetPostResponse {
  option (grpc.federation.message) = {};
  optional int64 score = 1 [(grpc.federation.field).by = "$.score"];
}
```

Reading the generated code, the request handoff at server.go.tmpl:546 uses req.GetScore(), which flattens *int64 to int64. So if a client omits score, $.score evaluates to 0, and the new ret.Score = &v wrap emits score: 0 as present — i.e. unset on the input side looks identical to score: 0 on the output side.

Is this the intended behavior for now? 
`_examples/23_optional` doesn’t cover this path, so I wasn’t sure whether it was out of scope or an oversight.

atkin-mc: This is intended behaviour for CEL. "$.score" on an unset optional returns the default value (0).

But this is great feedback -- because indeed, there is no way in the code generator to make an optional response field `nil` (i.e. don't set it). That will need fixing, but I will consider that in a separate PR.

generator/code_generator.go
atkin-mc: This is very good feedback, thank you. Ineed, this is a bug. I've followed your suggestions for refactoring.

resolver/resolver.go
atkin-mc: Moved up. Thanks. I don't see any way for the type to be nil, but this ordering doesn't make sense.

utahta (approved):

PR #363, opened 2026-03-03
Optimize WASM plugin execution with compilation caching

goccy:
- Optimize WASM plugin execution with compilation caching
- If result.Files does not exist, we will not run the code generator plugin

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -673,6 +673,9 @@ func CreateCodeGeneratorResponse(ctx context.Context, req *pluginpb.CodeGenerato
 	}
 
 	var resp pluginpb.CodeGeneratorResponse
+	// TODO: Since we don’t currently support editions, we will comment it out.
+	// Strictly speaking, proto3 optional is also not fully supported, but because it cannot be used together when other plugins support proto3 optional,
+	// we have enabled it for the time being.
 	resp.SupportedFeatures = proto.Uint64(uint64(pluginpb.CodeGeneratorResponse_FEATURE_PROTO3_OPTIONAL /*| pluginpb.CodeGeneratorResponse_FEATURE_SUPPORTS_EDITIONS*/))
 
 	if len(result.Files) == 0 {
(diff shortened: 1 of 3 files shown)


generator/generator.go
```diff
@@ -664,14 +671,23 @@ func CreateCodeGeneratorResponse(ctx context.Context, req *pluginpb.CodeGenerato
 		}
 		fmt.Fprint(os.Stderr, validator.Format(outs))
 	}
+	if len(result.Files) == 0 {
+		return nil, nil
```
utahta: Since returning nil can make protoc treat the plugin as not supporting FEATURE_PROTO3_OPTIONAL, should we return an empty response with SupportedFeatures set?

generator/wasm.go
```diff
@@ -3,37 +3,67 @@ package generator
...
 )
 
-func evalCodeGeneratorPlugin(ctx context.Context, pluginFile []byte, req io.Reader) (*pluginpb.CodeGeneratorResponse, error) {
+// wasmPlugin holds a compiled WASM module and its runtime, allowing
+// the module to be instantiated multiple times without recompilation.
+// Execute is safe for concurrent use.
+type wasmPlugin struct {
+	runtime  wazero.Runtime
+	compiled wazero.CompiledModule
+	counter  atomic.Int64
```
atkin-mc: Can you add a comment explaining what this counter is used for, please?

utahta (approved):

generator/generator.go
```diff
@@ -665,20 +672,30 @@ func CreateCodeGeneratorResponse(ctx context.Context, req *pluginpb.CodeGenerato
 		fmt.Fprint(os.Stderr, validator.Format(outs))
 	}
 
+	var resp pluginpb.CodeGeneratorResponse
+	resp.SupportedFeatures = proto.Uint64(uint64(pluginpb.CodeGeneratorResponse_FEATURE_PROTO3_OPTIONAL /*| pluginpb.CodeGeneratorResponse_FEATURE_SUPPORTS_EDITIONS*/))
```
atkin-mc, line 679: Is this commented-out part needed? If it is, could you add a comment to explain when it should be uncommented, please?

atkin-mc (changes requested):

atkin-mc (approved):

PR #365, opened 2026-04-13
Fix import alias confusion when multiple proto files share the same go_package

atkin-mc:
## Summary

Fixes a code generator bug where aliases are not resolved correctly when a Go package contained multiple `.proto` files.

This PR fixes code_generator.go `Imports` to make sure that different imported files that belong to the same (proto) Go package are assigned the same alias.


utahta (approved):
LGTM.

This fix is minimal, the updated `alias_multifile` test covers both the golden diff and a real `go build`, and it addresses a real-world breakage caused by multi-file packages.

One thought on the root cause — I am not fully confident in this analysis yet, so please take it as something worth investigating rather than a concrete proposal:

`resolver.resolvePackageAndFileReference` appears to create a fresh `*GoPackage` per file. If that is the case, then when two `.proto` files share the same `go_package`, they would end up with distinct pointers even though they represent the same Go package. A few code paths that key on pointer identity might therefore be exposed:

- `generator/code_generator.go` `pkgMap` / `aliasMap` — what this PR fixes.
- `resolver/message.go` — `base != msg.GoPackage()` could potentially treat same-package deps as distinct.
- `resolver/message.go` and `resolver/service.go` — pointer-keyed sets may not dedupe those packages as intended.
- `grpc/federation/generator/decode.go` — the plugin decode path might reintroduce the same hazard.

If this reading is correct, one possible direction worth exploring would be to canonicalize `*GoPackage` by `(Name, ImportPath)` in the resolver, and mirror that in the plugin decoder. That might also let us simplify or remove the `importPathToAlias` fallback introduced here, though I haven't verified it.

Either way, I would not bundle that into this PR. It could affect `GoPackageDependencies` output and would likely require revalidating the expected generated files, so it deserves its own investigation.

Issue #366, opened 2026-04-27
WASM CEL plugin mutex serialization causes per-process latency cliff under fan-out load


shiwork:
## Problem Statement

`grpc/federation/cel/plugin.go:739-750` serializes all WASM CEL plugin calls through a single `sync.Mutex`. Since `sync.Mutex.Lock()` is not context-aware, parked goroutines do not free their resources on upstream `ctx.Done()`, leading to memory accumulation under fan-out load.

## Reproduction Conditions

A service whose response message fans out to N data submessages, each invoking the same CEL function (e.g. a JWT parsing helper), generates N times the in-flight mutex acquisitions per request. At sufficient RPS, the per-process queue wait at `i.mu.Lock()` exceeds the upstream gateway's buffering deadline, requests time out, and the gateway re-queues retries while the original goroutines remain parked holding stack + closure + trace span memory until they eventually succeed (still doomed) or OOM occurs.

## Concrete Production Incident (anonymized)

- 1 RPC fans out to ~11 data submessages, 7+ of which call the same CEL function (a pure in-memory JWT parse with no I/O — i.e., a function that conceptually does not need WASM isolation).
- Production fan-out at ~500 RPS produced 14-34 second `Call` span queue waits, exceeding an upstream 25 s buffering deadline.
- Memory grew from steady state ~4 GB to ~5.5 GB / 6 GB pod limit before OOM.
- Horizontal scale-out (HPA min 20 → 40) did not help: per-pod mutex queue is unaffected by adding pods, and new pods were OOM-killed during warmup as concentrated traffic acquired their fresh mutex.

## Why context cancellation does not help

`sync.Mutex.Lock()` is not a context-aware primitive. There is no `LockContext(ctx)` in the standard library. Once a goroutine calls `Lock()`, it parks on `runtime_SemacquireMutex` and `ctx.Done()` cannot wake it. Resources held by parked goroutines (stack, closure, function arguments, trace span opened before the lock at `plugin.go:740`) accumulate linearly with queue depth. The framework's `gcWaitTimeout = 10 s` GC mitigation actually amplifies this because GC holds `i.mu` for up to 10 s.

## Proposed Discussion

Two distinct issues are interleaved here:

1. **No injection hook for native Go CEL function bindings.** Even when a CEL function is a pure CPU operation that does not need WASM isolation, the framework offers no path to register it as native Go code. All CEL functions registered via the plugin system pay the WASM dispatch + per-process mutex cost.
[...]

PR #368, opened 2026-04-27
feat(cel): make CELPluginInstance mutex context-aware

shiwork:
## Summary

Replaces `sync.Mutex` with `golang.org/x/sync/semaphore.Weighted(1)` on `CELPluginInstance`. Lock/Unlock pairs become `Acquire(ctx, 1)` / `Release(1)`, making the serialization point context-aware so upstream `ctx.Done()` actually unblocks parked goroutines.

Refs #366

## Motivation

The current `sync.Mutex` in `CELPluginInstance.Call` (`grpc/federation/cel/plugin.go`) serializes all WASM CEL plugin invocations per process. `sync.Mutex.Lock()` is not context-aware: a goroutine parked on it cannot be woken by `ctx.Done()`. Under load, this means:

- Request-scoped goroutines pile up at the lock
- Each parked goroutine holds its stack, function arguments (token bytes, metadata maps, CEL `ref.Val` slices), and the trace span (opened before the lock at `trace.Trace(ctx, "Call")`)
- Upstream cancellation (gateway timeout, client disconnect) cannot free them
- Memory pressure grows linearly with queue depth → OOM

A downstream service hit this in production on 2026-04-27. See #366 for the full incident background.

## Approach

Behavior preserved:
- Capacity-1 semaphore = same serialization as before
- `gcWaitTimeout = 10s` skip-on-timeout behavior preserved in the GC path (uses `context.WithTimeout(context.Background(), gcWaitTimeout)`)
- Backup-instance switchover semantics unchanged
- `Close()` uses `context.Background()` for Acquire so shutdown still waits for any in-flight call (same as the previous `sync.Mutex.Lock()` semantic on this path)

What changes:
- `Call` returns `ctx.Err()` immediately when cancelled while waiting for the instance, freeing the goroutine + closure + span resources
- `ValidatePlugin` is also context-aware now (it goes through the same lock)
- `startGC` bounds its wait at `gcWaitTimeout`: if a call is in flight longer than that, GC skips this tick (the next `enqueueGC` retries)
- `Release(1)` is only called on the success path of `Acquire` (per `golang.org/x/sync/semaphore` docs: "On failure, returns ctx.Err() and leaves the semaphore unchanged")

## Tests

- Existing tests: pass (`go test -race ./...`, including the WASM-based `TestPlugin` and `_examples/15_cel_plugin`, `_examples/21_wasm_net`)
- New tests in `grpc/federation/cel/plugin_test.go`:
  - `TestCallReleasesOnCtxCancel`: a Call blocked on the semaphore returns `context.Canceled` within 2s of `cancel()`
  - `TestCallReleasesOnCtxDeadline`: deadline propagation via `context.WithTimeout(50ms)`
  - `TestValidatePluginReleasesOnCtxCancel`: same context-aware behavior on the validation path
  - `TestStartGCSkipsOnAcquireTimeout`: confirms the GC code skips (returns `nil`) when it cannot acquire within `gcWaitTimeout`, without forcing the cancel/backup-switchover path

## Backward compatibility

- Public API unchanged
- Existing user code using `Call(ctx, ...)` sees identical behavior, except cancellation now actually works
- No proto / config schema change
- `golang.org/x/sync` is already an indirect dependency (`go.mod` v0.14.0); this PR does not change `go.mod` / `go.sum`

## Relationship to #367 and #366

- This PR addresses one of the two issues raised in #366 (the mutex-not-context-aware property)
- It is independent of #367 (`AdditionalCELOptions` injection hook): the two PRs touch disjoint files and can merge in any order
- The orthogonal "pool of N instances" approach was considered but rejected for this PR because it changes the implicit single-instance contract for plugin authors. This PR is a strictly conservative change (same serialization, just cancellable)

## Risks

- Goroutines that previously blocked indefinitely on the mutex (waiting for an `Unlock`) now exit early on ctx cancel. Code that depended on the synchronous "Call completes regardless of ctx" semantics would change behavior. I am not aware of any such consumer; in normal gRPC server flow, ctx cancel is the expected exit signal

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/grpc/federation/cel/plugin.go
+++ b/grpc/federation/cel/plugin.go
@@ -22,6 +22,7 @@ import (
 	"github.com/tetratelabs/wazero"
 	"github.com/tetratelabs/wazero/api"
 	"github.com/tetratelabs/wazero/imports/wasi_snapshot_preview1"
+	"golang.org/x/sync/semaphore"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
@@ -414,7 +415,11 @@ func (p *CELPlugin) createInstance(ctx context.Context) (*CELPluginInstance, err
 		gcQueue:          make(chan struct{}, gcQueueLength),
 		gcErrCh:          make(chan error, 1),
 		instanceModErrCh: make(chan error, 1),
-		cancelFn:         cancel,
+		// sem serializes invocations into the WASM instance with capacity 1
+		// (equivalent to sync.Mutex), but Acquire is context-aware so that
+		// upstream cancellation can free parked goroutines.
+		sem:      semaphore.NewWeighted(1),
+		cancelFn: cancel,
 	}
 	instance.id = fmt.Sprintf("%s-%p", p.cfg.Name, instance)
 
@@ -566,9 +571,15 @@ type CELPluginInstance struct {
 	instanceModErr   error
 	gcErrCh          chan error
 	closed           bool
-	mu               sync.Mutex
-	gcQueue          chan struct{}
-	cancelFn         context.CancelFunc
+	// sem is a capacity-1 weighted semaphore that serializes calls into the
+	// underlying WASM instance, replacing what was previously a sync.Mutex.
+	// Acquire(ctx, 1) is context-aware: when the caller's context is canceled
+	// while waiting, Acquire returns ctx.Err() and the goroutine (along with
+	// its stack, function arguments, and trace span) can be released, instead
+	// of remaining parked indefinitely as it would on sync.Mutex.Lock().
+	sem      *semaphore.Weighted
+	gcQueue  chan struct{}
+	cancelFn context.CancelFunc
 }
 
 const PluginProtocolVersion = 2
@@ -589,8 +600,12 @@ func (i *CELPluginInstance) ValidatePlugin(ctx context.Context) error {
 	ctx, span := trace.Trace(ctx, "ValidatePlugin")
 	defer span.End()
 
-	i.mu.Lock()
-	defer i.mu.Unlock()
+	if err := i.sem.Acquire(ctx, 1); err != nil {
+		// Acquire returns ctx.Err() on cancellation and leaves the semaphore
+		// unchanged, so Release must NOT be called on this path.
+		return err
+	}
+	defer i.sem.Release(1)
 
 	if err := i.write(ctx, []byte(VersionCommand)); err != nil {
 		return fmt.Errorf("failed to send cel protocol version command: %w", err)
@@ -645,8 +660,14 @@ func (i *CELPluginInstance) Close() error {
 		return nil
 	}
 
-	i.mu.Lock()
-	defer i.mu.Unlock()
+	// Close is called during shutdown and must wait for any in-flight call to
+	// complete before tearing down resources. Acquire with a Background ctx
+	// blocks until success and never returns an error, preserving the
+	// previous sync.Mutex.Lock() semantic on this path.
+	if err := i.sem.Acquire(context.Background(), 1); err != nil {
+		return err
+	}
+	defer i.sem.Release(1)
 
 	return i.close()
 }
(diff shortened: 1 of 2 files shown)


grpc/federation/cel/plugin.go
atkin-mc, line 714: Could you please split into two separate timeouts? (Both can be 10s). e.g.

// timeout to acquire semaphore for GC
var gcWaitForSemTimeout = 10 * time.Second
// timeout for a GC run
var gcRunUntilTimeout = 10 * time.Second

grpc/federation/cel/plugin.go
```diff
@@ -697,8 +718,22 @@ func (i *CELPluginInstance) startGC(ctx context.Context) error {
 	ctx, span := trace.Trace(ctx, "github.com/mercari/grpc-federation.CELPluginInstance.startGC")
 	defer span.End()
 
-	i.mu.Lock()
-	defer i.mu.Unlock()
+	// GC is not bound to any user request context. Use a fresh
+	// Background-derived ctx with gcWaitTimeout so that if the semaphore is
```
atkin-mc, line 722:
```suggestion
	// Background-derived ctx with gcWaitForSemTimeout so that if the semaphore is
```

atkin-mc: (and below)

grpc/federation/cel/plugin.go
atkin-mc, line 742:
```suggestion
	ctx, cancel := context.WithTimeout(ctx, gcRunUntilTimeout)
```

grpc/federation/cel/plugin.go
```diff
@@ -697,8 +718,22 @@ func (i *CELPluginInstance) startGC(ctx context.Context) error {
...
+	// this GC attempt rather than blocking the GC goroutine forever. The
+	// previous sync.Mutex.Lock() could not time out, so this is also a
+	// strict improvement on top of context-awareness.
+	acquireCtx, acquireCancel := context.WithTimeout(context.Background(), gcWaitTimeout)
+	defer acquireCancel()
+	if err := i.sem.Acquire(acquireCtx, 1); err != nil {
+		// Acquire timed out (or the ctx was canceled). Skip this GC tick;
+		// the next enqueueGC tick will retry. Do NOT call Release here:
+		// semaphore is unchanged on Acquire failure.
+		log.Logger(ctx).WarnContext(ctx, fmt.Sprintf("grpc-federation: skipping GC because instance lock could not be acquired within %s: %v", gcWaitTimeout, err))
```
atkin-mc, line 733: It might be worth adding a span attribute here so the skip path is discoverable in APM 
trace search without needing a log-string alert, e.g.
```     
span.SetAttributes(trace.StringAttr("acquire_status", "cancelled"))             
```
or maybe
```
status := "canceled"
if errors.Is(err, context.DeadlineExceeded) {
  status = "timeout"                                                       
}                                                                               
```
If it's useful to have more signal, you can also do
```
span.SetStatus(codes.Error, err.Error())                                      
span.RecordError(err)
```

grpc/federation/cel/plugin.go
```diff
@@ -740,8 +775,17 @@ func (i *CELPluginInstance) Call(ctx context.Context, fn *CELFunction, md metada
...
-	defer i.mu.Unlock()
+	// Acquire is context-aware. If ctx is canceled (upstream gateway timeout
+	// or client disconnect) while waiting for the in-flight call to finish,
+	// Acquire returns ctx.Err() and this goroutine — along with its stack,
+	// arguments, and the trace span opened above — can be released, instead
+	// of staying parked on a non-cancellable sync.Mutex.Lock().
+	if err := i.sem.Acquire(ctx, 1); err != nil {
+		// Per semaphore.Weighted docs: "On failure, returns ctx.Err() and
+		// leaves the semaphore unchanged." Do NOT call Release on this path.
+		return nil, err
```
atkin-mc, line 786: Also perhaps `span.SetAttributes(..)` here, too. But since this is the more frequent code path I don't think it's worth doing `span.SetStatus` here -- it will potentially trigger on every call to the plugin (not just 1 GC call per BFF request).

grpc/federation/cel/plugin.go
atkin-mc, line 284: With your change, this is going to make a timeout calling the plugin turn into a GRPC `codes.Unknown` error, when it would be better to have Canceled or DeadlineExceeded. I will make a separate PR to fix this part.

atkin-mc (changes requested):
Thank you for this PR. It looks good overall, but I have made a few comments.

I will fix the `NewErrFromString` issue with error propagation, in a separate PR.

atkin-mc:
Based on this PR, I created https://github.com/mercari/grpc-federation/pull/369. It's reviewed and merged now.

PR #370, opened 2026-05-11
Allow setting an optional field to nil using field.by

atkin-mc:
PR https://github.com/mercari/grpc-federation/pull/361 added support for proto3 optional fields, but it did not provide a way to leave them unset (i.e. a non-nil value must be assigned even to an optional field). This PR changes the semantics for `grpc.federation.field.by` so that, if the value of the CEL expression is `optional.none()`, the field is left unset, instead of the former behaviour of setting it to the "zero value" for the field.

This change only affects CEL values for `grpc.federation.field.by` when the field is optional -- non-optional fields are not affected. Since optional field support is not released yet, it is a non-breaking change.

Scope: Only field.by on proto3 optional scalar/enum fields is affected. Proto3 optional message and bytes fields already represent absence as nil via their Go types and are unchanged. Oneof fields, def variables, and other binding mechanisms are out of scope.


grpc/federation/cel.go
```diff
@@ -863,29 +872,35 @@ func EvalCEL(ctx context.Context, req *EvalCELRequest) (any, error) {
...
 	out, _, err := program.ContextEval(ctx, evalValues)
 	if err != nil {
-		return nil, err
+		return nil, false, err
 	}
 	opt, ok := out.(*celtypes.Optional)
 	if ok {
 		if opt == celtypes.OptionalNone {
-			return reflect.Zero(req.OutType).Interface(), nil
+			if req.OutType == nil {
```
atkin-mc, line 884: The change to check output type is nil is for consistency with non-optional case -- this check is missing. According to control flow in the code generator, output type is never nil, but checking to be safe against future changes.

utahta (approved):

PR #372, opened 2026-06-08
Allow CEL libraries to also be defined externally

atkin-mc:
Allow CEL libraries to be defined outside grpc-federation (an "external CEL library") and imported by the compiler. Type information continues to be supplied using the CEL plugin definition mechanism `(grpc.federation.file).plugin.export`. A CEL library defined this way is treated in exactly the same way as a library defined in `grpc/federation/cel`. It is up to the service initialization code to ensure that the set of `cel.SingletonLibrary` instances supplied at startup time matches the CEL plugin definitions that were supplied to the compiler, including the library name and function declarations.

The motivation for this change is to provide a lightweight way for externally-defined CEL functions to be supplied to the CEL runtime. External CEL libraries are semantically identical to WASM plugins, but without the compile overhead of WASM and the performance impact of calling into the WASM sandbox.

A service can use *either WASM plugins or external libraries*, but not both. See `_examples/24_cel_library` for an example of usage.

* n.b. This PR intentionally splits `resolver/resolver_cel_library_test.go` out from the very large `resolver/resolver_test.go`, as a first step to splitting the latter into smaller files.


generator/generator.go
```diff
@@ -622,7 +625,10 @@ func (g *Generator) generateByGRPCFederation(r *PluginRequest) (*pluginpb.CodeGe
 	relativePath := g.absPathToRelativePath[r.protoPath]
 	pathResolver := resolver.NewOutputFilePathResolver(opt.Path)
 
-	result, err := resolver.New(r.req.GetProtoFile(), resolver.ImportPathOption(opt.Path.ImportPaths...)).Resolve()
+	result, err := resolver.New(r.req.GetProtoFile(),
+		resolver.ImportPathOption(opt.Path.ImportPaths...),
+		resolver.CELLibrariesOption(g.cfg.CELLibraries...),
```
utahta, line 639: When Config.CELLibraries is set, should we force `protoc-gen-grpc-federation` to use the in-process path before [checking cfg.installedPath](https://github.com/mercari/grpc-federation/blob/b07a007c362dc08e66faebbd4da7fd7b10220da7/generator/generator.go#L279-L294)? The external plugin cannot receive these Go library values through CodeGeneratorRequest. What do you think?

_examples/24_cel_library/cmd/codegen/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+	r := resolver.New(files,
+		resolver.ImportPathOption("proto"),
+		resolver.CELLibrariesOption(extlib.NewLibrary()),
+	)
+	result, err := r.Resolve()
+	if err != nil {
+		return err
+	}
+	if len(result.Files) != 1 {
```
utahta: This command assumes result.Files contains exactly one file, but this example can also include the plugin.export file in the resolver result. That makes `cmd/codegen` fail before writing the generated service. Selecting the resolved file that contains the federation service, instead of requiring `len(result.Files) == 1`, should fix this.
e.g.:
```go
  var targetFile *resolver.File
  for _, file := range result.Files {
        if len(file.Services) != 0 {
                targetFile = file
                break
        }
  }
  if targetFile == nil {
        log.Fatal("federation service file not found")
  }
```

generator/templates/server.go.tmpl
```diff
@@ -257,7 +259,7 @@ func New{{ $serviceName }}(cfg {{ $serviceName }}Config) (*{{ $serviceName }}, e
 	}
 	{{- end }}
 	{{- if $celPlugins }}
-	if cfg.CELPlugin == nil {
+	if cfg.CELPlugin == nil && len(cfg.CELLibraries) == 0 {
 		return nil, grpcfed.ErrCELPluginConfig
```
utahta, line 263: The ErrCELPluginConfig message still says "CELPlugin field is not set. this field must be set", but the generated config gate now accepts either cfg.CELPlugin or cfg.CELLibraries.
Should we update the error message to mention both alternatives, e.g. "CELPlugin or CELLibraries field must be set"?

generator/generator.go
atkin-mc: Yes, you are right about this, thanks very much for raising it.

I think this is theoretically possible (can pass type & functions defintions through CodeGeneratorRequest), but for the time being it is better to just fall back to the in-process plugin.

_examples/24_cel_library/cmd/codegen/main.go
atkin-mc: Good catch, thanks!

utahta (approved):

generator/templates/server.go.tmpl
atkin-mc: Fixed

